PR #15, opened 2017-10-07
fix(dependecy) update babel to bable-cli

MurhafSousli:
fix npm installation failure because `babel` is deprecated in favor of `babel-cli`

capture


adrianq:
@MurhafSousli I am merging this as, as you said, it makes more sense to use babel-cli but are you sure this hasn't happened to you due to a global npm installation?

PR #16, opened 2017-10-07
Feature (Filter by)

### LINKED ISSUE #5: Add filters
- [x] Add filter by user roles
- [x] Add filter by user group
- [x] Modify search textfield to look into the rest of the columns

MurhafSousli:
closes #5


MurhafSousli:
@adrianq PR is ready for reviewing

adrianq:
Ok, there is a clear misunderstanding here. I am gonna try to explain myself again however it may be worth that we have a quick chat to clarify any doubt. The task is about:
- Keeping the old search textfield that will search in all columns always. This textfield will always look in all columns and its behaviour is always the same and independent of the other textfields.
- Add two dropdowns. One with user roles and another one with user groups.  These dropdowns will be populated with all the user roles and user groups existing in the DHIS2 instance. Users will be able to select one of the options. Let's say the user select a user role in the user role dropdown filter. Let's say he chooses SO_dataset_education. The application will show only the users that have this user role. Same applies for user group.
- You can filter by a user group (selecting one of the user group in the user groups dropdown) and then search by a string in the search textfield. Let's say we select in the dropdown GL_CampManagement and the app will display only the users that belong to this user group. Then the user search for Adriana in the textfield and the app will show all the users that belong to GL_CampManagement and that have the string "Adriana" in any field.
- I am aware the list of user roles and user groups is huge and therefore the dropdown will be showing a very large list of options. We will probably change this into a more convenient component but for now, let's use dropdowns for user roles and user groups.
Let me know if you have any doubt and we can talk tomorrow...

MurhafSousli:
@adrianq Ok, now I got it

src/List/list.store.js
```diff
@@ -65,30 +89,55 @@ export default Store.create({
...
         getD2().then(d2 => {
             if (!d2.models[modelType]) {
                 error(`${modelType} is not a valid schema name`);
             }
-
             let modelDefinition;
             if (searchString) {
-
-                if (searchBy === 'name') {
+                if (filterByRole && filterByGroup) {
```
MurhafSousli: @adrianq I have tried couple of ways to chain the filtering query, but I couldn't get the filter query to work properly, do you think this filtering syntax has problem?

MurhafSousli:
@adrianq This PR is ready but I have a question regard filtering

adrianq:
As we discussed, the filter and rootJunction parameters is not working properly so we will perform a js filtering.
For future reference: https://docs.dhis2.org/2.27/en/developer/html/webapi_metadata_object_filter.html

MurhafSousli:
@adrianq PR is ready

tokland:
I've kept most of Murhaf code, the most important changes affect the userRoles/userGroups filters. Filtering in JS space forces a reimplementation of the paginator, which is not trivial. I've confirmed (see #5) that there is a bug for dhis2 api < 2.30 for multiple filters on nested fields, what I do now: individual requests on those two filters, get the user IDs and use them to build the final filtering request (text/canManage).

PR #54, opened 2018-04-20
Use custom feedback-tool with dhis2/github module

tokland:
Closes https://github.com/EyeSeeTea/dataset-configuration/issues/159


adrianq:
I merged it but the screenshot feature is not working particularly good. I guess we have too many react components... We can have a look at the other lib in the future

PR #60, opened 2018-05-24
Fetch translations with same-origin credentials to avoid 302

### LINKED ISSUE #59: Translations not working on feedback-tool
302 Login on en.properties request

Applies also to ds-config

tokland:
Closes #59 

When getting translations with `fetch`, we must use `credentials: 'include'` so it works on production.

Documentation:
https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#Sending_a_request_with_credentials_included


adrianq:
Working fine now! See http://dev.eyeseetea.com:8080/api/apps/user-app/index.html I am gonna patch the previous release and update the zip file

PR #67, opened 2018-07-27
Replicate users dialog from a table (A2.2)

### LINKED ISSUE #66: Replicate user: create from table (A.2.2)
See documentation, A2.2

Fields: User will type the username (compulsory), full name, email and password (compulsory)

tokland:
### :pushpin: References

* **Issue:** close #66 
* **Required PRs:** https://github.com/EyeSeeTea/d2-ui/pull/23
* **Merge after PRs:** #65 

### :tophat: What is the goal?

An admin can replicate a user using a table to specify username/password (required) and firstName/surname/email (keep the values from user if empty). Note that the proposal document specified a `fullName` field, but DHIS2 DB uses `firstName` and `surname`. For now, I've kept 2 fields. Alternative:  Keep single column, but split fields (requires changes in d2-ui)

### :memo: How is it being implemented?

- Added support for sub-items in contextual actions.
- Added support to d2-ui/FormBuilder so the tag wrapper can be passed as props. This way we can directly use the FormBuilder inside the table using wrappers `TableRow` and `TableRowColumn`. See https://github.com/EyeSeeTea/d2-ui/pull/23.
- Added `ReplicateUserFromTable` React component.

### :art: Screenshots

* Contextual action with subitems:

screenshot from 2018-07-27 13-33-37

* Example table with 2 users (max 100):

screenshot from 2018-07-27 13-34-11


tokland:
Sorry, I had some of those changes already committed, but not the pushed :( I am leaving it here so tomorrow I can confirm, but your changes look good: http://download.zaudera.com/public/changes.diff

adrianq:
Perfect, works like a charm!

PR #81, opened 2018-08-30
Market setup

### LINKED ISSUE #79: Prepare app for market
- [x] Allows defining header colour based on manifest variable. Two options for now: NRC & general
- [x] Rename app from user-app to user-extended-app
- [x] Add Eyeseetea logo on the right of feedback tool. Suggestions: https://drive.google.com/file/d/0B298kwCziFFQM3kxcklFV0gxWEU/view?usp=sharing Add variable in manifest to enable/disable this component.

tokland:
### :pushpin: References

* **Issue:** Closes #79 
* Builds on PR #80 

### :memo: Implementation

- Added keys `appKey` and `appearance` to the app-config.json files. [Example.](https://github.com/EyeSeeTea/user-app/blob/feature/market-setup-79/app-config.example.json)
- Name: User App -> User Extended App
- Added logo. Temporal, let's see if Ray has suggestions.
- **Important note about the app key**: In 2.26, if you installed an app from `my-app.zip`, that would create a folder `apps/my-app/` and the app would have key `my-app`, regardless of its real name. However, from 2.28, the app key is always the app name with dashes: "Some dhis2 app" -> "Some-dhis2-app". In our case: "User-Extended-App". That's why we have this appKey in the app config files. To sum it up: it should just work for >= 2.28. For < 2.28, make sure the zip has the same name of  appKey, this way we don't have to change the configuration file.  To see the info of installed apps, this works within user-app:  `d2.system.installedApps`.

### :art: Screenshots

screenshot from 2018-08-30 11-40-35


adrianq:
@tokland I have made a few changes. Most of them are in [the last commit](https://github.com/EyeSeeTea/user-app/pull/81/commits/9b84f29c6f1a401fb7d4f0515bd78092c9d9ac10). Can you please have a look at the them and merge if you are happy with them?

tokland:
Looks good, just some minor code refactors. I've added also hover styles for the button. I've moved the event arrow functions to avoid re-renders.

PR #113, opened 2018-10-29
Refactor/polish

### LINKED ISSUE #112: Polish after deploy
- [x] On import -> file select -> show spinner at the very start.
- [x] On import start, /orgUnits load takes a lot, use filter by displayName:in, also for userRoles and userGroups.
- [x] Review translation key (missing: "override existing user") [@adrianq]
- [x] Trash icon disappears in some cases. [@adrianq]
- [x] Limit number of warnings in import table.
- [x] Cut imported rows to maxUsers and add warning.
- [x] Some users cannot be deleted (username_8) -> 500 -> APPARENTLY, DB PROBLEM, NOTHING TO DO.
- [x] Show also username in warning -> `Value not found: ${missingValue} [username=${username} csv-row=${rowIndex} column=${modelName}]`

https://docs.google.com/spreadsheets/d/1i4ZXX5SpAa1MTsplvdcOAnUlp_WiP2cZMPKWed3K8Cg/edit?ts=5bd2e56f#gid=0

tokland:
### :pushpin: References

* **Issue:** Closes #112 

### :memo: Implementation

- There was no .babelrc in the project, after some clean up on global node_modules, I started to got some errors. I've added the same we are using in dataset-configuration.
-  The sub-task "On import start, /api/orgUnits takes a lot to load, use filter=displayName:in" has not been trivial, since d2-ui orgUnits components return only the path of selected orgUnits and now we don't have the full info. Refactored the component, which is also used in 1) filters, 2) orgUnits assign contextual actions.
- I've set onChange events of text fields onBlur, so only when focus is lost, model data is updated.
- A couple of subtasks are pending, all yours, @adrianq


adrianq:
@tokland I have pushed the changes regarding the deployment context path. Can you please double check with your local changes if they are equivalent? These ones seem to work fine locally and in a remote server. I couldn't reproduce the delete icon issue but I found this minor issue https://github.com/EyeSeeTea/user-extended-app/issues/119. Updated as well the excel sheet with some comments. Merging this now...

PR #132, opened 2018-11-26
Add checkbox show-only-active

### LINKED ISSUE #131: Add filter: show only active users
- [x] Move "display only users i can manage" + new checkbox inside extended filters box.
- [ ] Implement "show only active users"  (check userCredentials.disabled)
- [ ] Add "Disabled" column

tokland:
### :pushpin: References

* **Issue:** Closes #131 

### :memo: Implementation

- The filters were in the List component and getting bigger, I've moved it to its own component: Filters.component.
- `DetailsBoxWithScroll.component.js` also extracted.
- Since we are filtering by the enabled/disable attribute, I've added it as column (same views -with a check- than official users app). It's not a sortable column (not supported by the API).
- Filter on `userCredentials.disabled` field.

### :art: Screenshots

screenshot from 2018-11-26 14-04-46


adrianq:
@tokland merging this as I would like to check the other PRs. However I would like to discuss with you the behaviour of the 'disabled' column on import/export. Disabled column is actually exported but values are always empty. Then disabled column is ignored on import. At least we should avoid exporting the disabled column even if it is displayed. Will try to talk to @ifoche re this so we have his input

tokland:
https://github.com/EyeSeeTea/user-extended-app/pull/142

PR #138, opened 2018-12-09
move reset inside advanced filters and minor cosmetics

adrianq:


tokland:
- [ ] I've moved in-line styles to the styles object
- [ ] The `X` button looks very nice inside the box, but note that it clear all filters, even the text filter outside the box.

PR #153, opened 2019-09-07
Add prettier configuration

SferaDev:
### Implementation

- Add the code style we now use in web apps to the user-extended app

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "d2-user-extended-app",
-  "version": "0.2.2",
+  "version": "0.2.3",
   "description": "DHIS2 Extended User app",
   "main": "src/index.html",
   "license": "GPL-3.0",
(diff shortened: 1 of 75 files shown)


package.json
```diff
@@ -19,6 +19,7 @@
     "build": "rm -rf build && webpack && npm run manifest",
     "postbuild": "cp -rv app-config.json src/i18n includes icon.png app-favicon.ico node_modules/babel-polyfill/dist/polyfill.min.js node_modules/jquery/dist/jquery.min.js ./build/",
     "build-webapp": "yarn build && rm -f $npm_package_name.zip && cd build && zip -r ../$npm_package_name.zip *",
+    "prettify": "prettier \"{src,scss}/**/*.{js,jsx,ts,tsx,json,css,scss}\" --write",
```
adrianq: @SferaDev What about adding the [code quality action](https://github.com/EyeSeeTea/vaccination-app/blob/development/package.json#L57) and the husky prepush hook?

SferaDev: On metadata-synchronization we never had this command!

The husky hook is okay but I didn't see the point most of the times it was executed other than adding extra time to the git push. As the changes were ignored in the push... 

I believe we should talk about this topic with the whole web team and streamline the behavior across projects. 

In any case, since these changes do not affect the code or the build we can leave postpone them for another release.

adrianq: Agreed. As discussed, I will just merge this PR as is and we can discuss about the hooks in the evaluation meeting

Issue #164, opened 2019-10-21
[GEN] Import is replacing OU assignment


SferaDev:
Client: WHO (General Maintenance)
Redmine: http://who-dev.essi.upc.edu:8083/issues/2699

When importing an user if the OU column is missing, the app is resetting the property to an empty array instead of MERGING with the old value.

This is important when importing changes that do not imply OU assignments and we do not want to export all the columns. (ie: User templates that are being regularly imported with role changes).

This probably happens not only with OU but other fields.

PR #166, opened 2019-10-27
Do not clear collections on import

SferaDev:
### Description

This was hard to debug, even though our logic manually merged the CSV built user with the existing user during import, we had a bug with collections (userGroups/userRoles/organisationUnits/dataViewOrganisationUnits) that were defaulting to an empty collection instead of undefined.

### Steps to reproduce

- Export some users without "Organisation Units" column
- Import these users

**Expected behavior**

- Users should have kept the original organisation unit assignment

**Actual behavior**

- Users have lost their assignment to the organisation unit

NOTE: The case where you have the column and no OU assigned has always worked

### DIFF
--- a/src/models/userHelpers.js
+++ b/src/models/userHelpers.js
@@ -58,8 +58,8 @@ const columnNameFromPropertyMapping = {
     created: "Created",
     userRoles: "Roles",
     userGroups: "Groups",
-    organisationUnits: "OUOutput",
-    dataViewOrganisationUnits: "OUCapture",
+    organisationUnits: "OUCapture",
+    dataViewOrganisationUnits: "OUOutput",
 };
 
 const propertyFromColumnNameMapping = _.invert(columnNameFromPropertyMapping);
--- a/src/models/userHelpers.js
+++ b/src/models/userHelpers.js
@@ -162,10 +162,12 @@ function collectionFromNames(user, rowIndex, field, objectsByName) {
             `Value not found: ${missingValue} [username=${username ||
                 "-"} csv-row=${rowIndex} csv-column=${field}]`
     );
-    const objects = _(objectsByName)
-        .at(names)
-        .compact()
-        .value();
+    const objects = objectsByName
+        ? _(objectsByName)
+              .at(names)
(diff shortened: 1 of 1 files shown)


adrianq:
@SferaDev This is not working for all the cases. Export a user without 'organisationUnits' and 'roles' columns, import again, roles are kept but organisationUnits are empty

adrianq:
Closes #164

SferaDev:
> @SferaDev This is not working for all the cases. Export a user without 'organisationUnits' and 'roles' columns, import again, roles are kept but organisationUnits are empty

That is strange, I have reproduced the same you are mentioning and it is working properly on my builds...

For example, I've just exported 10 users only with username, first name and surname and imported it again:

image

Both organisation unit assignment and roles are now kept instead of cleared (merging properly works).

And the call to metadata import contains the values, even though they have not been provided in the CSV:

image

And if we compare the importedFields with the baseFields we see that the collections are not defined anymore.

image
image

@adrianq If you can please verify the patch is in your build and/or provide an CSV file to reproduce on my end I could test it with your scenario.

adrianq:
Sure @SferaDev , I have double-checked it right now, still failing:
1) Select just one single user (Asavela_Shabangu). I am testing in dev2 8083
2) Export without organisation units and roles. I have checked  the user actually has organisation units and roles 
image
3) Modify surname and append a '2' at the end to be sure the import was successful
4) Modify table to display organisation units and roles
5) Execute the import
6) Table is automatically refreshed. Roles are still there but organisation units have disappeared. 

This test case is quite consistent so you should be able to reproduce it if you perform exactly the same steps. Anyway, let me know if you want me to jump into a hangout.

SferaDev:
Ah I see now the problem, including OUCapture but no OU is the scenario failing because the two cases check the same array.

image

Since OUCapture array look-up did not fail, OU is getting the data origin from that collection.

I have fixed the issue by adding an extra condition in this PR that verifies the field value does not actually exist in the imported CSV.

I have replicated the behavior and now it does not fail.

PR #184, opened 2020-12-11
Copy user configuration to other users

### LINKED ISSUE #172: Copy user configuration to other users
- [x] When clicking on "Copy User, show a modal dialog
- [x] Dialog should contain two toggles "user groups" and "user roles"
- [x] Implement logic to copy user groups
- [x] Add logic to copy user groups and/or user roles

-------------------------------
- [x] Add new contextual action "Copy in User"
- [x] Action should open a modal dialog
- [x] Show a multiple selector in the dialog with the available users in the DHIS2
- [x] On save, copy the user roles to the selected users. We should merge the user roles from the origin with the user roles of the destination users.

-------------------------------
For the future:

Tracking: http://who-dev.essi.upc.edu:8083/issues/3376

Allow to copy configuration from a user as a template to multiple existing users.

Use-case: Template user "WIDP Admin" applied to new members.
    - [ ] Allows to select: OU/CaptureOU
    - [ ] Replace/Merge options

jvfdunkley:
### :pushpin: References

-   **Issue:** Closes #172

### :memo: Implementation

- I created a new Dialog component called CopyInUserDialog where I edited the getPayload method. The getPayload method takes a zipped pair consisting of the parentUser and the child(ren)Users and for each child user, it either copies over all of the parent's userRoles using _m.imerge if the child had no userRoles at all, or it adds userRoles that the parent has but the child doesn't
- In BatchModelsMultiSelect.component, I added a copyInUser boolean state variable that the CopyInUserDialog could pass in so on the "SAVE" button click, it routes to the copyInUserSave method (which is async) instead of the standard save method. 
- In BatchModelsMultiSelect.model I added the copyInUserSave method as well as a getUserInfo helper method which calls the '/users' GET endpoint to get all the info for a user. I called this for both the parent and the children and then passed them into the payload function and then finally POSTed the resulting object. 

### :art: Screenshots
Screen Shot 2020-12-11 at 11 06 17 AM

### :fire: Testing
I have accounted for the case of a childUser having no userRoles and there being multiple childUsers selected. I have not written any tests because user-extended-app doesn't have any relevant tests but I can if requested.


src/components/CopyInUserDialog.component.js
```diff
@@ -0,0 +1,67 @@
...
+}
+
+function getPayload(pairs) {
+    //I need to do this because when the pair maps, there might be more than 1 childUser but only 1 parentUser so
+    //when it iterates the 2nd time for the other child, there is no 2nd parentUser
+    const parent = pairs[0][0]
+    const users = pairs.map(([user, childrenUsers]) => {
+        let childrenUserRoles = childrenUsers.userCredentials.userRoles
+        if(childrenUserRoles.length == 0)
+        {
```
tokland, line 20: New/updated code is not prettified. `yarn prettify" does it, but I'd recommend to enable the option "Format on save" in vscode, so you don't have to worry about it, you'll see it's really practical.

src/components/CopyInUserDialog.component.js
```diff
@@ -0,0 +1,67 @@
...
+import BatchModelsMultiSelectComponent from "./batch-models-multi-select/BatchModelsMultiSelect.component";
+import _m from "../utils/lodash-mixins";
+
+function getTitle(getTranslation, users) {
+    const usernames = users && users.map(user => user.userCredentials.username);
+    const info = usernames ? _m.joinString(getTranslation, usernames, 3, ", ") : "...";
+    return getTranslation("copyInUser") + ": " + info;
+}
+
+function getPayload(pairs) {
```
tokland, line 13: Logic functions should be in helpers (usersHelper or a new one you create). The idea is to keep code in the components relevant only to React.

src/components/CopyInUserDialog.component.js
```diff
@@ -0,0 +1,67 @@
...
+                userCredentials: _m.imerge(childrenUsers.userCredentials, { userRoles: parent.userCredentials.userRoles}),})
+        }
+        else
+        {
+            parent.userCredentials.userRoles.forEach(role => {
+                if(childrenUserRoles.find(element => element.id == role.id) == undefined)
+                {
+                    childrenUserRoles.push(role)
+                }
+            })
```
tokland, line 31: I know there is a lot to process as it is, so not for now, but we favor functional code (no in-place updates). Some docs here: https://github.com/EyeSeeTea/know-how/wiki/Typescript

src/components/CopyInUserDialog.component.js
```diff
@@ -0,0 +1,67 @@
...
+    const modelOptions = {
+        parentModel: d2.models.users,
+        parentFields: ":owner,userCredentials[id,username,userRoles[id,name],lastLogin]",
+        childrenModel: d2.models.users,
+        getChildren: user => user.userCredentials.userRoles,
+        getPayload: getPayload,
+    };
+
+    return (
+        <BatchModelsMultiSelectComponent
```
tokland, line 51: You have integrated the funcionality in BatchModelsMultiSelectComponent. For now we don't use the merge/update features of that component, so another alternative would have been creating a separate component. We can keep it as-it-is.

src/components/CopyInUserDialog.component.js
```diff
@@ -0,0 +1,67 @@
...
+function getTitle(getTranslation, users) {
+    const usernames = users && users.map(user => user.userCredentials.username);
+    const info = usernames ? _m.joinString(getTranslation, usernames, 3, ", ") : "...";
+    return getTranslation("copyInUser") + ": " + info;
+}
+
+function getPayload(pairs) {
+    //I need to do this because when the pair maps, there might be more than 1 childUser but only 1 parentUser so
+    //when it iterates the 2nd time for the other child, there is no 2nd parentUser
+    const parent = pairs[0][0]
```
tokland, line 16: If you have to write `pairs[0][0]`, something should be changed. A function should have clean arguments. Here it looks like we need `function getPayload(parentUser, destUsers) { ...`. And then, change the code where getPayload is being called.

tokland (commented):
The feature works well! Some comments about the code:

tokland (approved):
I've re-tested the feature (works well) and pushed some minor refactors, take a look at them. Otherwise, as far as I am concerned, the PR is good to go!

adrianq:
Already tested as part of https://github.com/EyeSeeTea/user-extended-app/pull/185

Issue #41, opened 2018-02-21
Fix lint/test


tokland:
This should work: npm test && npm run lint

(Remove or fix)

adrianq:
Not relevant anymore as the architecture is quite old

PR #190, opened 2021-02-09
Workaround for paging error when filtering by group

jvfdunkley:
### :pushpin: References

-   **Issue:** Closes ["Paging error filtering by group"](https://app.clickup.com/t/dn075t) in ClickUp  and issue [#41](https://github.com/EyeSeeTea/user-extended-app-blessed/issues/41)

### :memo: Implementation 
- The issue was that in the API call if there was no searchString in the query it would only return the first page of results instead of all results so I just changed it so that if there is a searchString you add the query param else don't

### :art: Screenshots
For this example now it shows all 60 users instead of stopping at 50, the page limit 
Screen Shot 2021-02-09 at 11 03 22 AM

Screen Shot 2021-02-09 at 10 56 35 AM

### :fire: Testing

### DIFF
--- a/src/List/Filters.component.js
+++ b/src/List/Filters.component.js
@@ -119,8 +119,9 @@ export default class Filters extends React.Component {
 
         const inFilter = field => (_(field).isEmpty() ? null : ["in", field]);
 
-        let returnObj = {
+        return {
             ...(showOnlyManagedUsers ? { canManage: "true" } : {}),
+            ...(searchString ? { query: searchString } : {}),
             filters: {
                 "userCredentials.disabled": showOnlyActiveUsers ? ["eq", false] : undefined,
                 "userCredentials.userRoles.id": inFilter(userRoles),
@@ -129,8 +130,6 @@ export default class Filters extends React.Component {
                 "dataViewOrganisationUnits.id": inFilter(orgUnitsOutput.map(ou => ou.id)),
             },
         };
-
-        return searchString ? { ...returnObj, query: searchString } : returnObj;
     }
 
     clearFilters = () => {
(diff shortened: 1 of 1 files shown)


src/List/Filters.component.js
```diff
@@ -119,8 +119,7 @@ export default class Filters extends React.Component {
 
         const inFilter = field => (_(field).isEmpty() ? null : ["in", field]);
 
-        return {
-            query: searchString,
+        let returnObj = {
```
tokland: The typical approach: use `let` for variables which are assigned more than once (example: a let + conditional with sets it on the different branches). For all other cases: `const`.

src/List/Filters.component.js
```diff
@@ -130,6 +129,8 @@ export default class Filters extends React.Component {
                 "dataViewOrganisationUnits.id": inFilter(orgUnitsOutput.map(ou => ou.id)),
             },
         };
+
+        return searchString ? { ...returnObj, query: searchString } : returnObj;
```
tokland: Not that this is bad, but, for consistency, we can replicate the idea used with `showOnlyManagedUsers`, it's a typical pattern to conditionally set a key/value in an object. The other way is to do something like `_.omitBy(myObject, _.isNil)`, which removes any property that has null/undefined values. That's possible only when those values are not valid for this particular object, of course. Note that _.isNil = _.isNull + _.isUndefined, we can use any other predicate we want.

tokland (approved):
Good point, as a rule-of-thumb, we should only send params which have a relevant value. I've just refactored the code a little bit, and left some comments.

That said, this solution is a workaround to what seems a DHIS2 bug. When making requests with query+filter, the pagination is off, wrongly limiting it to 50 records. This existing bug seems related. It says it's fixed for 2.31/2.32, but something very similar is happening to us:

https://jira.dhis2.org/browse/DHIS2-8956

PR #191, opened 2021-02-11
Add strategy to copy in user feature. 

jvfdunkley:
### :pushpin: References

-   **Issue:** Closes ["Add strategy to copy in user feature"](https://app.clickup.com/t/dn0840) in ClickUp

### :memo: Implementation 
- Added Replace/Merge toggle in top right and moved property toggles to bottom left
- In getPayload in userHelpers I passed the updateStrategy chosen to the addItems function so if it is chosen as "Replace" then it will just return the parent version of that property. 
- I then had to create a sort of new version of saveUsers in saveCopyInUsers where it got the existing users so it overrode their values.

### :art: Screenshots
Screen Shot 2021-02-11 at 3 38 29 PM

### :fire: Testing

### DIFF
--- a/src/components/batch-models-multi-select/CopyInUserBatchModelsMultiSelect.component.js
+++ b/src/components/batch-models-multi-select/CopyInUserBatchModelsMultiSelect.component.js
@@ -201,21 +201,6 @@ export default class CopyInUserBatchModelsMultiSelectComponent extends React.Com
             />,
         ];
     }
-    /*
-    <Toggle
-                    label={"Merge"}
-                    style={this.styles.strategyToggle}
-                    checked={this.state.copyUserGroups == true}
-                    onToggle={(ev, newValue) => this.setState({ copyUserGroups: newValue })}
-                />
-                <Toggle
-                    label={"Replace"}
-                    style={this.styles.strategyToggle}
-                    checked={this.state.copyUserRoles === true}
-                    onToggle={(ev, newValue) => this.setState({ copyUserRoles: newValue })}
-                />
-
-    */
 
     render() {
         const isLoading = this.state.state === "loading";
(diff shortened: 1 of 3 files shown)


src/components/batch-models-multi-select/CopyInUserBatchModelsMultiSelect.component.js
```diff
@@ -105,25 +101,20 @@ export default class CopyInUserBatchModelsMultiSelectComponent extends React.Com
...
-            return null;
-        }
+        const label =
+            this.getTranslation("update_strategy") +
+            ": " +
+            this.getTranslation("update_strategy_" + this.state.updateStrategy);
+        return (
+            <Toggle
+                label={label}
+                style={{ width: 280, float: "right", marginTop: 20, marginRight: 15 }}
```
tokland, line 111: [minor, no need to change] As a rule of thumb, we use immutable props. For styles, that means using the syles object with a reference here.

src/models/userHelpers.js
```diff
@@ -638,43 +647,50 @@ async function getExistingUsers(d2, options = {}) {
...
+
+function addItems(items1, items2, shouldAdd, updateStrategy) {
     if (!shouldAdd) {
         return items1;
     } else {
-        return _(items1)
-            .unionBy(items2, element => element.id)
-            .value();
+        return updateStrategy === "merge"
+            ? _(items1)
```
tokland, line 656: [very minor, no need to change] When only one lodash function if used, the non-wrapped version: `_.unionBy(...)` uses only one line. The pro of doing it this is is that it's very easy to add more lodash methods to the chain.

src/models/userHelpers.js
```diff
@@ -638,43 +647,50 @@ async function getExistingUsers(d2, options = {}) {
...
+                  .unionBy(items2, element => element.id)
+                  .value()
+            : items2;
     }
 }
 
-function getPayload(parentUser, destUsers, fields) {
+function getPayload(parentUser, destUsers, fields, updateStrategy) {
```
tokland, line 663: [minor, no need to change] When a function starts growing, a typical refactor is to use a single options argument object. This way, the call to the function has each field named, which helps readability.

tokland (approved):
:+1: Looks good to me! Just some minor in-line comments, but no need to implement them, just as general ideas:

PR #194, opened 2021-02-18
Check user extended in 2.35 - Move CSS dependencies from a remote URL to inside the project and fix import error

jvfdunkley:
### :pushpin: References

-   **Issue:** Closes ["Check user extended in 2.35"](https://app.clickup.com/t/e35bxx) in ClickUp

### :memo: Implementation 
- Copied font-awesome, fonts and material-design-icons folders from dhis-core repository to the project and put it in a new css folder in the project root
- edited the webpack config so that the CSS files will just reference the root folders in the build folder
- added to the postbuild command in package.json where it copies over various folders to the build folders so that the font-awesome, fonts and material-design-icons folders are there and can be referenced by the CSS files
- Fixed an import error due to an API call? Currently investigating

### :art: Screenshots
The navigation arrows show up properly so that's how you know it works. Also there should ideally be no MIME type errors in the console.
Screen Shot 2021-02-18 at 2 10 09 PM

### :fire: Testing


tokland (commented):
CSS and fonts working at dev/build, cool. Tested all the features of the app, I only found some errors when trying to save users with references to user groups that the current user cannot save, which is normal.

My only doubt is about having the font binaries statically in the app. Before: 4.6M, After: 6.2M. @adrianq Not sure if it's worth the effort to generate instead css with references to CDN servers. We have 3 parts for this: font-awesome, roboto fonts, material UI icons.

tokland (approved):
As discussed, the only gain of using CDN fonts would be to the reduce the ZIP size, but the user experience would be equivalent. So let's keep it this way.

PR #196, opened 2021-03-04
download empty user template

jvfdunkley:
### :pushpin: References

-   **Issue:** Closes ["Download empty template for importing users"](https://app.clickup.com/t/ewz19t) in ClickUp

### :memo: Implementation 
- I added an option in the import/export dropdown to export an empty template
- I passed all columns to the ImportExport component and mapped through it for their labels

### :art: Screenshots
Screen Shot 2021-03-04 at 9 52 33 AM
Screen Shot 2021-03-04 at 9 53 00 AM

### DIFF
--- a/src/List/List.component.js
+++ b/src/List/List.component.js
@@ -91,7 +91,7 @@ const List = React.createClass({
 
     mixins: [ObserverRegistry, Translate, Auth],
 
-    maxImportUsers: 1000,
+    maxImportUsers: 200,
 
     styles: {
         dataTableWrap: {
--- a/src/List/List.component.js
+++ b/src/List/List.component.js
@@ -91,7 +91,7 @@ const List = React.createClass({
 
     mixins: [ObserverRegistry, Translate, Auth],
 
-    maxImportUsers: 200,
+    maxImportUsers: 1000,
 
     styles: {
         dataTableWrap: {
@@ -508,6 +508,7 @@ const List = React.createClass({
                         <ImportExport
                             d2={d2}
                             columns={settings.get("visibleTableColumns")}
+                            allColumns={allColumns}
                             filterOptions={listFilterOptions}
                             onImport={this._openImportTable}
                             maxUsers={this.maxImportUsers}
--- a/src/components/ImportExport.component.js
+++ b/src/components/ImportExport.component.js
@@ -75,7 +75,7 @@ class ImportExport extends React.Component {
         }
     };
 
-    exportEmptyTemplate = async () => {
+    exportEmptyTemplate = () => {
         const { allColumns } = this.props;
         this.setState({ isProcessing: true });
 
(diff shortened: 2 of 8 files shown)


src/components/ImportExport.component.js
```diff
@@ -75,6 +75,23 @@ class ImportExport extends React.Component {
...
+            const datetime = moment().format("YYYY-MM-DD_HH-mm-ss");
+            const filename = `empty-user-template-${datetime}.csv`;
+            FileSaver.saveAs(blob, filename);
+            snackActions.show({ message: `${this.t("table_exported")}: ${filename}` });
+        } finally {
+            this.closeMenu();
+            this.setState({ isProcessing: false });
+        }
+    };
+
```
tokland, line 92: The common code with `exportToCsvAndSave` should be abstracted somehow. There are two typical approaches: 1) abstract the full method passing the particularities to the function (even on the form of functions, it behavious must be abstracted), or 2) leave them more independent and abstract the pieces (more repetition but more versatile and usually more clear). There are pros and cons for all approaches. Check bfbc7bf for my (2) approach, I've abstracted only the specific `saveCsv` code.

src/components/MultipleSelector.component.js
```diff
@@ -81,7 +81,6 @@ class MultipleSelector extends React.Component {
         organisationUnitsCapture: "assignToOrgUnitsCapture",
         dataViewOrganisationUnits: "assignToOrgUnitsOutput",
     };
-
```
tokland: [very minor] I've revert this change, we typically don't commit this changes if the file was not changed for the task (it's ok to tidy up files in the general case, though)

src/List/List.component.js
```diff
@@ -508,6 +508,7 @@ const List = React.createClass({
                         <ImportExport
                             d2={d2}
                             columns={settings.get("visibleTableColumns")}
+                            allColumns={allColumns}
```
tokland, line 511: This new prop was missing in PropTypes, added: 30701d3. Fortunately, we don't need this for TS apps.

src/components/ImportExport.component.js
```diff
@@ -119,15 +142,20 @@ class ImportExport extends React.Component {
...
                             primaryText={t("export")}
                             onClick={exportToCsvAndSave}
                         />
                         <MenuItem
-                            leftIcon={<ImportIcon />}
+                            leftIcon={<ExportIcon />}
                             primaryText={t("import")}
                             onClick={importFromCsv}
                         />
+                        <MenuItem
```
tokland, line 160: It's a bit weird to have this order (Export/Import/Export empty), but I am not aware of the requirements, so I'll leave @adrianq to decide.

src/models/userHelpers.js
```diff
@@ -266,7 +266,6 @@ async function getUsersFromCsv(d2, file, csv, { maxUsers, orgUnitsField }) {
         : _(csv.data)
               .drop(1)
               .value();
-
```
tokland: Same

tokland (approved):
All well! Just some minor comments and commits I made:

src/components/ImportExport.component.js
adrianq: Yes, I agree I have just changed the order to Import, Export and Export Empty

PR #197, opened 2021-03-05
add disable column in spreadsheet

jvfdunkley:
### :pushpin: References

-   **Issue:** Closes ["Add disable column in spreadsheet"](https://app.clickup.com/t/ewz22j) in ClickUp

### :memo: Implementation 
- I added the disabled column in ImportExport
- For the export function, I just had to pass in all columns instead of filtering by the ones that are not supposed to be shown on export
- For the import I added disabled to the userCredentials field so that it would be copied to the right place when updating a user
- The disabled value would show up as a string when it had to be a boolean so I had to convert it in the getUsersFromCsv function
 
### :art: Screenshots
Screen Shot 2021-03-05 at 11 09 15 AM
Screen Shot 2021-03-05 at 11 23 48 AM

### DIFF
--- a/src/models/userHelpers.js
+++ b/src/models/userHelpers.js
@@ -61,7 +61,7 @@ const columnNameFromPropertyMapping = {
     userGroups: "Groups",
     organisationUnits: "OUCapture",
     dataViewOrganisationUnits: "OUOutput",
-    disabled: "disabled",
+    disabled: "Disabled",
 };
 
 const propertyFromColumnNameMapping = _.invert(columnNameFromPropertyMapping);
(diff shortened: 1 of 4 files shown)


tokland (changes requested):

src/models/userHelpers.js
```diff
@@ -266,7 +266,6 @@ async function getUsersFromCsv(d2, file, csv, { maxUsers, orgUnitsField }) {
         : _(csv.data)
               .drop(1)
               .value();
-
     const plainUserAttributes = _(d2.models.users.modelValidations)
         .map((value, key) => (_(["TEXT", "DATE", "URL"]).includes(value.type) ? key : null))
```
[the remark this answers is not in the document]
adrianq: @tokland I would say we can render the boolean columns as a toggle switch. Can you take care of this @jvfdunkley?

src/models/userHelpers.js
```diff
@@ -337,7 +336,9 @@ async function getUsersFromCsv(d2, file, csv, { maxUsers, orgUnitsField }) {
         const data = userRows.map((userRow, rowIndex) =>
             getPlainUserFromRow(userRow, modelValuesByField, rowIndex + 2)
         );
-        const users = data.map(o => o.user);
+        const users = data.map(o =>
+            o.user.disabled ? { ...o.user, disabled: o.user.disabled.toLowerCase() } : o.user
```
tokland, line 340: I'd keep it a boolean, something like: `disabled: o.user.disabled === "true"`

tokland, line 340: Sorry, I hadn't notice before, but this processing for the field disabled could be in "getPlainUserFromRow", right?

tokland (changes requested):
Some in-line comments. Check with @adrianq, but I think we need some extra work on the import part. In the import table, show a toggle, and make sure a boolean value (not a string) is posted.

tokland (approved):
I pushed some changes for the toggle, works for me!

PR #198, opened 2021-03-16
Paging does not refresh properly when removing filtered roles

jvfdunkley:
### :pushpin: References

-   **Issue:** Closes ["Paging does not refresh properly when removing filtered roles"](https://app.clickup.com/t/f555gn) in ClickUp

### :memo: Implementation 
- It took about 10 seconds for the pagination to actually return the correct value in the API call so it must be a cache issue on the API side or something else 
- As a workaround, in list.store, filter function I made a second getList call without pagination so it returned all users that fit the filter
- Then created a buildPager helper function which takes the list of users with pagination and the second getList variable. This function made the pager.total equal the length of the users Map and then returned it so the total number would actually reflect the number of users that are returned.
 
### :art: Screenshots
Not a UI difference so it is hard to show. It was 442 and I deleted a user and now it is 441
Screen Shot 2021-03-16 at 4 03 16 PM

### DIFF
--- a/src/List/List.component.js
+++ b/src/List/List.component.js
@@ -448,8 +448,8 @@ const List = React.createClass({
 
     render() {
         if (!this.state.dataRows) return null;
+        //const { pager } = this.state;
         const currentlyShown = calculatePageValue(this.state.pager);
-        const { pager } = this.state;
         const { d2 } = this.context;
 
         const paginationProps = {
@@ -458,10 +458,14 @@ const List = React.createClass({
             hasPreviousPage: () =>
                 Boolean(this.state.pager.hasPreviousPage) && this.state.pager.hasPreviousPage(),
             onNextPageClick: () => {
-                this.setState({ isLoading: true }, () => this.filterList({ page: pager.page + 1 }));
+                this.setState({ isLoading: true }, () =>
+                    this.filterList({ page: this.state.pager.page + 1 })
+                );
             },
             onPreviousPageClick: () => {
-                this.setState({ isLoading: true }, () => this.filterList({ page: pager.page - 1 }));
+                this.setState({ isLoading: true }, () =>
+                    this.filterList({ page: this.state.pager.page - 1 })
+                );
             },
             total: this.state.pager.total,
             currentlyShown,
--- a/src/List/List.component.js
+++ b/src/List/List.component.js
@@ -448,7 +448,6 @@ const List = React.createClass({
 
     render() {
         if (!this.state.dataRows) return null;
-        //const { pager } = this.state;
         const currentlyShown = calculatePageValue(this.state.pager);
         const { d2 } = this.context;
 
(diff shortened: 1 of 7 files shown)


src/List/list.store.js
```diff
@@ -95,11 +95,29 @@ export default Store.create({
...
+    },
     filter(options, complete, error) {
         getD2().then(d2 => {
             const { filters, ...listOptions } = options;
-            const listSearchPromise = getList(d2, filters, listOptions);
-            this.listSourceSubject.onNext(Observable.fromPromise(listSearchPromise));
+            const listSearchPromiseWithPagination = getList(d2, filters, listOptions);
+            const listSearchPromiseWithAllUsers = getList(d2, filters, {
+                ...listOptions,
+                paging: false,
```
tokland, line 112: As we only need the users count for this second request, I'd ask for fewer fields. This should work here: `fields: "id",`

src/List/List.component.js
```diff
@@ -458,10 +457,14 @@ const List = React.createClass({
...
-                this.setState({ isLoading: true }, () => this.filterList({ page: pager.page + 1 }));
+                this.setState({ isLoading: true }, () =>
+                    this.filterList({ page: this.state.pager.page + 1 })
+                );
             },
             onPreviousPageClick: () => {
-                this.setState({ isLoading: true }, () => this.filterList({ page: pager.page - 1 }));
+                this.setState({ isLoading: true }, () =>
+                    this.filterList({ page: this.state.pager.page - 1 })
+                );
```
tokland, line 467: I guess this change was no really needed for the feature to work, but you wanted to keep a homogeneous usage of `this.state.pager` in all functions, right?

src/List/list.store.js
```diff
@@ -95,11 +95,29 @@ export default Store.create({
         this.listSourceSubject.onNext(Observable.fromPromise(this.state.pager.getPreviousPage()));
     },
 
+    buildPager(listSearchPromiseWithPagination, listSearchPromiseWithAllUsers) {
+        return Promise.all([listSearchPromiseWithPagination, listSearchPromiseWithAllUsers]).then(
+            ([model, model1]) => {
```
tokland, line 100: [minor] proposal: `[paginatedUsers, allUsers]`

src/models/user.js
```diff
@@ -76,7 +75,6 @@ class User {
             users: userGroup.users.concat(newUsers.map(newUser => ({ id: newUser.id }))),
         }));
         const payload = { users: newUsers, userGroups: userGroupsWithNewUsers };
-
```
tokland, line 79: [minor] Blank lines on non-relevant code.

tokland (changes requested):
It works, that's a weird behaviour in his we hadn't notice before

src/List/list.store.js
```diff
@@ -95,11 +95,29 @@ export default Store.create({
...
+                model.pager.total = model1.valuesContainerMap.size;
```
tokland, line 101: Unfortunately, this is not enough. The next/prev page handling may be also off. We seem to use `getNextPage`, `getPreviousPage`, we should override them (they are functions that return a promise, check the original implementation for reference)

tokland (approved):

PR #199, opened 2021-04-14
Fix error when searching in the textfield and more than 50 matches are found

jvfdunkley:
### :pushpin: References

-   **Issue:** Closes ["Fix error when searching in the textfield and more than 50 matches are found"](https://app.clickup.com/t/gv5kx1) in ClickUp

### :memo: Implementation 

- The issue was that params `filter` and `query` couldn't be in the same request. The filter was automatically set to "name:ne:default" - filter all names that don't equal the default user's name. 
-
- The solution was to not include `filter` instead of having that default setting so the query wouldn't be overridden or messed up.

### :art: Screenshots
Before - only 50 would show up when there are 270 or so matches
screenshot-1617806230161698532
After - now it shows all 270 when you query "z"
Screen Shot 2021-04-14 at 2 40 40 PM

### DIFF
--- a/src/List/Filters.component.js
+++ b/src/List/Filters.component.js
@@ -118,6 +118,7 @@ export default class Filters extends React.Component {
         } = this.state;
 
         const inFilter = field => (_(field).isEmpty() ? null : ["in", field]);
+
         return {
             ...(showOnlyManagedUsers ? { canManage: "true" } : {}),
             ...(searchString ? { query: searchString } : {}),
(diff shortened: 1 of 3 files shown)


tokland (commented):
We checked this together, I didn't notice that this alleviates the problem, but we still need to support query+filter (when search by text and filtering by role or group, for example), so this won't be enough. I think the conclusion is that the DHIS2 pager for the users endpoint cannot be trusted and we should simply rebuild it, as we do in https://github.com/EyeSeeTea/user-extended-app/pull/198. I think we can just close this PR without merging, and get the fix from #198. Not that merging this will hurt, it makes no sense to have this "name:ne:default" for users.

PR #200, opened 2021-05-14
OU Capture not showing up and can't click on GUI when replicating user

jvfdunkley:
### :pushpin: References

-   **Issue:** Closes ["Bug: OU Capture not showing up and can't click on GUI when replicating user"](https://app.clickup.com/t/vd8pa0) in ClickUp

### :memo: Implementation 
- Made the OrgUnitForm show the correct labels for OU capture and OU output
- Made the translations all like org_unit_output instead of orgUnitOutput
- The MultiplerSelector component recognizes the orgUnits and dataViewOrgUnits now since I switched it back so the GUI popup will appear on click
- Unfortunately I couldn't switch all the organisation_units to org_units_capture and data_view_organisation_units to organisation_units_output in the codebase because I think the translations are used directly in the models and the user API uses organisationUnit and dataViewOrganisationUnit so I could not change that aspect. So the translations use the same variable on the backend for consistency but all relevant translations will be OU Capture and OU Output on the frontend. 

### :art: Screenshots
Screen Shot 2021-05-14 at 2 27 28 PM


src/components/OrgUnitForm.js
```diff
@@ -174,7 +182,7 @@ class OrgUnitForm extends React.Component {
                 <TextField
                     onChange={event => this._searchOrganisationUnits(event.target.value)}
                     floatingLabelText={this.context.d2.i18n.getTranslation(
-                        "filter_organisation_units_capture_by_name"
+                        `${filteringByNameLabel}`
```
tokland, line 185: No need to interpolate. Same below.

src/i18n/i18n_module_es.properties
```diff
@@ -1,13 +1,13 @@
...
 assign_all=Assignar todos
-assignGroups=Asignar a grupos
-assignRoles=Asignar a roles
-assignToOrgUnitsCapture=Asignar a unidades organizativas
-assignToOrgUnitsOutput=Asignar a unidades organizativas de salida
-replicateUser=Replicar usuario
+assign_groups=Asignar a grupos
+assign_roles=Asignar a roles
+assign_to_org_units_capture=Asignar a la captura de unidades organizativas
+assign_to_org_units_output=Asignar a la salida de las unidades de la organización
```
tokland, line 6: Not sure about "unidades de la organización", we have "unidades organizativas" in the previous translation.

src/List/organisation-unit-dialog/OrgUnitDialog.component.js
```diff
@@ -144,6 +144,16 @@ class OrgUnitDialog extends React.Component {
                     roots={this.props.roots}
                     selected={this.state.selected}
                     intersectionPolicy={true}
+                    filteringByNameLabel={
+                        title.includes("organisation units capture")
```
tokland, line 148: This kind of checks are very brittle, whenever we change the literal it will stop working. I'd add these two props to the component and the caller must pass them. 

[minor] Also, not sure about passing the translation key as prop. Maybe just the already translated string?

tokland (changes requested):

tokland (approved):
All good!

PR #209, opened 2021-10-13
[Refactor] Fix Copy In User New

jvfdunkley:
### :pushpin: References

-   **Issue:** Part 1 of ?? in [[User-Extended] Review all contextual actions if they work or not, fix old code to make them work again](https://app.clickup.com/t/1fzw77u) in ClickUp. Fixing the Copy In User functionality.

### :memo: Implementation 
- Now the users are passed as an array so I modified the PropTypes and took out user destructuring in CopyInUserDialog.component.js since that is not necessary anymore
- There was a mismatch between OrgUnits and OrgUnitOutputs. I thought in the previous fixes in user-extended-app we changed the names to be OU Capture and OU Output. 
- Parent is passed as an string id now instead of object in CopyInUserBatchModelsMultiSelect.component.js and CopyInUserBatchModelsMultiSelect.model.js so I removed some lodash things to get the ids.
- Improved typing of filters from previous PR. I created a Filter type. 

### :art: Screenshots
No visual change

### DIFF
--- a/src/legacy/components/batch-models-multi-select/CopyInUserBatchModelsMultiSelect.component.js
+++ b/src/legacy/components/batch-models-multi-select/CopyInUserBatchModelsMultiSelect.component.js
@@ -1,8 +1,6 @@
 import LoadingMask from "d2-ui/lib/loading-mask/LoadingMask.component";
+import { ConfirmationDialog } from "@eyeseetea/d2-ui-components";
 import _ from "lodash";
-import Dialog from "material-ui/Dialog/Dialog";
-import FlatButton from "material-ui/FlatButton/FlatButton";
-import RaisedButton from "material-ui/RaisedButton/RaisedButton";
 import TextField from "material-ui/TextField/TextField";
 import Toggle from "material-ui/Toggle/Toggle";
 import PropTypes from "prop-types";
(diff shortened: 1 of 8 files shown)


src/legacy/components/batch-models-multi-select/CopyInUserBatchModelsMultiSelect.component.js
```diff
@@ -211,14 +188,15 @@ export default class CopyInUserBatchModelsMultiSelectComponent extends React.Com
...
-                autoDetectWindowHeight
-                contentStyle={this.styles.dialog}
+                maxWidth={"lg"}
+                fullWidth={true}
                 open={true}
-                onRequestClose={this.props.onRequestClose}
+                onCancel={this.props.onCancel}
+                cancelText={this.getTranslation("cancel")}
+                infoActionText={this.getTranslation("save")}
+                onInfoAction={!isLoading ? this.copy.bind(this) : undefined}
```
SferaDev, line 199:
```suggestion
                saveText={this.getTranslation("save")}
                onSave={!isLoading ? this.copy.bind(this) : undefined}
```

@jvfdunkley For future PRs

PR #210, opened 2021-10-14
Write and define roadplan for clean architecture rewrite

jvfdunkley:
### :pushpin: References

-   **Issue:**  [[User-Extended] Write and define roadplan for clean architecture rewrite](https://app.clickup.com/t/1fzw7jn) in ClickUp. Fixing the Copy In User functionality.


src/domain/entities/Settings.ts
```diff
@@ -0,0 +1,5 @@
+export interface Settings {
+    visibleTableColumns: string[];
+    organisationUnitsField: organisationUnitsFieldSelect;
```
SferaDev, line 3: Globally-set setting

src/domain/entities/Settings.ts
```diff
@@ -0,0 +1,5 @@
...
+}
+type organisationUnitsFieldSelect = "shortName" | "code";
```
SferaDev, line 5:
```suggestion
type OrganisationUnitsFieldSelect = "shortName" | "code";
```

PR #213, opened 2021-10-18
[Refactor] Fix remove user 

jvfdunkley:
### :pushpin: References

-   **Issue:** Part 4 of ?? in [[User-Extended] Review all contextual actions if they work or not, fix old code to make them work again](https://app.clickup.com/t/1fzw77u) in ClickUp. Fixing the remove user functionality.

### :memo: Implementation 
- For some reason the snackbar onActionTouchTap wasn't working at all and I couldn't figure out why. I replaced it with a confirmationDialog since that is the intended usecase of the component. 
- Since I was using the confirmationDialog I changed the userListTable action so it sets the state of the deleteUserStore so I can show the confirmationDialog and then onConfirm I call the deleteStore.delete function which does the API call. I removed the performDelete function called in the snackAction and merged it with the normal delete function.
- Add translations for confirmationDialog title

### :art: Screenshots
Screen Shot 2021-10-18 at 10 16 01 PM

### DIFF
--- a/public/old-i18n/i18n_module_ar.properties
+++ b/public/old-i18n/i18n_module_ar.properties
@@ -127,7 +127,6 @@ extended_filters=Advanced filters
 clear_filters=Clear all filters
 table_layout_at_least_one_column=Select at least one column
 confirm=Confirm
-remove_users=Remove user(s)?
 confirm_delete_users=Are you sure you want to remove the selected users? ($$users$$)
 users_deleted=Users removed successfully: $$users$$
 config_import_export=Import/export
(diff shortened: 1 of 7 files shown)


public/old-i18n/i18n_module_ar.properties
```diff
@@ -127,6 +127,7 @@ extended_filters=Advanced filters
 clear_filters=Clear all filters
 table_layout_at_least_one_column=Select at least one column
 confirm=Confirm
+remove_users=Remove user(s)?
```
SferaDev, line 130: For new translations, we should use the new i18n instead of the old one.

Also the dialog title can be ``Remove users`` without the question mark

SferaDev (changes requested):

PR #217, opened 2021-11-07
QA fixes / ConfirmationDialog switch

oventi:
### :pushpin: References
-   **Issue:** https://app.clickup.com/t/1eqy1u7 and https://app.clickup.com/t/1jvcwha

__IMPORTANT__
Another task that is modifying the ImportTable, and that component uses MultipleSelector, which uses the old Dialog and OrgUnitSelector component. These files are not modified and will need to be at a later time.

### :memo: Implementation 
- Fixed "undefined" on edit user
- Added Api url in the details box
- Added bigger paging options
- Fied lastLogin view message
- Fixed button clock when replicating users
- Added OrgUnitsSelectorFilter component with helper functions to keep the selection functionality
- Switched to ConfirmationDialog instead of Dialog

### :art: Screenshots

qa_fixes_1

qa_fixes_2

qa_fixes_3

qa_fixes_4

qa_fixes_5


jvfdunkley:
@oventi please follow the format for [our PRs](https://github.com/EyeSeeTea/user-extended-app/pull/216) including the task name and link in ClickUp. Good to keep things consistent and clear, even if Arnau and Alexis sometimes do their own thing ;)

src/legacy/components/ReplicateUserFromTemplate.component.js
```diff
@@ -57,7 +57,8 @@ class ReplicateUserFromTemplate extends React.Component {
     };
 
     onUpdateFormStatus = formStatus => {
-        const isValid = !formStatus.asyncValidating && formStatus.valid;
+        const asyncValidating = "asyncValidating" in formStatus && formStatus.asyncValidating;
```
jvfdunkley, line 60: We shouldn't use "property" in "object". I forget exactly why Alexis didn't want to use it, maybe for performance. I think instead you should use lodash mapValues and see if the key is there. @SferaDev can chime in if this is actually an ok situation to use property in object and not mapValues. 

Here's an example of code that I corrected from sharing-settings-app when I wanted to originally do `if('visualizations' in payload)`

`function extractExtraDependencies(payload: MetadataPayload): string[] {`
  `  return _(payload)`
       ` .mapValues((value, key) => {`
           ` if (key === "visualizations") {`
               ` return _.flatten(`
                   ` value.map((element: Visualization) =>`
                       ` _.flatMap(element.dataDimensionItems ?? [], (item: DataDimensionItem) => {`
                            `const indicator = item.indicator?.id;`
                            `const programIndicator = item.programIndicator?.id;`
                           ` return _.compact([indicator, programIndicator]);`
                       ` })`

jvfdunkley (approved):
I have that one comment that I'm not sure is correct and needs Alexis' input but overall looks good! I tested everything myself and it all works as intended. Good job :)

src/domain/entities/User.ts
```diff
@@ -10,11 +10,12 @@ export interface User {
     email: string;
     lastUpdated: Date;
     created: Date;
+    apiUrl: string;
     userRoles: UserRole[];
     userGroups: NamedRef[];
     organisationUnits: NamedRef[];
     dataViewOrganisationUnits: NamedRef[];
-    lastLogin: Date;
+    lastLogin: Date | string;
```
SferaDev, line 18: In domain ``lastLogin`` should be a ``Date``, when parsing/encoding to JSON (data repository implementation) dates should be properly formatted (either manually or with a schema model)

src/data/repositories/UserD2ApiRepository.ts
```diff
@@ -61,8 +63,9 @@ export class UserD2ApiRepository implements UserRepository {
             created: new Date(user.created),
             userGroups: user.userGroups,
             username: user.userCredentials.username,
+            apiUrl: `${this.api.baseUrl}/api/users/${user.id}.json`,
             userRoles: user.userCredentials.userRoles,
-            lastLogin: new Date(user.userCredentials.lastLogin),
+            lastLogin: lastLogin === undefined ? "N/A" : new Date(lastLogin),
```
SferaDev, line 68: Instead of setting here ``N/A`` for undefined values, this should be handled on the ``presentation`` layer in a component scope.

src/data/repositories/UserD2ApiRepository.ts
```diff
@@ -51,6 +51,8 @@ export class UserD2ApiRepository implements UserRepository {
     }
 
     private mapUser(user: D2ApiUser): User {
+        const lastLogin = user.userCredentials.lastLogin;
```
SferaDev, line 54: Unless we re-use/transform a variable we normally use them directly instead of copying them to a local variable.

src/legacy/List/context.actions.js
```diff
@@ -31,8 +31,9 @@ function lexicographicalCompare(xs, ys) {
     return _(xs).zipWith(ys, compare).find() || 0;
 }
 
-export async function goToUserEditPage(user) {
+export async function goToUserEditPage(userId) {
     const d2 = await getD2();
+    const user = (await d2.models.users.list({ filter: `id:in:[${userId}]` })).toArray().pop();
```
SferaDev, line 36: Not sure about this ``.pop`` method. What is the array returning?

src/legacy/components/OrgUnitsSelectorFilter.js
```diff
@@ -41,44 +38,33 @@ class OrgUnitsFilter extends React.Component {
...
         this.setState({ selected });
     }
 
-    applyAndClose = () => {
-        this.props.onChange(this.state.selected);
-        this.closeDialog();
-    };
+    async applyAndClose() {
+        const { d2 } = this.context;
+        const orgUnitIds = getOrgUnitsPaths(this.state.selected).map(path => _.last(path.split("/")));
```
SferaDev, line 59: The path to id conversion is heavily used, maybe a helper method instead of repeating the last/split combination could be helpful.

```
export function cleanOrgUnitPath(orgUnitPath?: string): string {
    return _(orgUnitPath).split("/").last() ?? "";
}

export function cleanOrgUnitPaths(orgUnitPaths: string[]): string[] {
    return orgUnitPaths.map(cleanOrgUnitPath);
}
```

src/legacy/components/ReplicateUserFromTemplate.component.js
SferaDev: Agreed, the ``property in object`` could be refactored to a more functional approach

src/legacy/utils/dhis2Helpers.js
```diff
@@ -10,6 +10,9 @@ function getOrgUnitsRoots() {
         .toPromise();
 }
 
+const getOrgUnitsPaths = ouOrPaths =>
+    ouOrPaths.map(ouOrPath => (typeof ouOrPath === "object" && "path" in ouOrPath ? ouOrPath.path : ouOrPath));
```
SferaDev, line 14: Same as before, we wouldn't use ``typeof`` or ``in operator`` here.

And, actually, I don't see the need for this helper method. We shouldn't mix string/object in a same array. So I'd rather check where this method is being used and transform to ``string[]`` before concatenating the items.

SferaDev (changes requested):
Thanks @oventi for the PR and @jvfdunkley for the review (please take a look at the comments I left for Andres too).

I've left some comments in the code, but they're minor, almost everything is good as-is, mostly code style that needs small adjustments.

Regarding the PR template, yes we should follow the same template as it simplifies changing context between apps/teams. Also we try to standardized a little bit the name of the branches ``feature/name`` ``fix/name`` and PR titles.

oventi:
Apologies, @jvfdunkley @SferaDev, I did not get the notifications. Will handle your feedback now.

oventi:
> @oventi please follow the format for [our PRs](https://github.com/EyeSeeTea/user-extended-app/pull/216) including the task name and link in ClickUp. Good to keep things consistent and clear, even if Arnau and Alexis sometimes do their own thing ;)

Thanks for reminding me. Don't know what happened, sometimes the templates does not appear, or maybe I overrode it 😄

src/legacy/components/ReplicateUserFromTemplate.component.js
oventi: Refactored to remove `in`

src/legacy/List/context.actions.js
oventi: I'm using `.pop` because the method `d2.models.users.list` returns a list of users and I only need 1. I'm sure I get one because I'm sending a user id. Do you prefer I use `[0]`?

src/legacy/components/OrgUnitsSelectorFilter.js
oventi: I'm not sure what you are asking, should I add the above code in helpers and then use it in line 59 and in the other places that use the same code? And which function?

src/legacy/utils/dhis2Helpers.js
oventi: This code was for this task: https://app.clickup.com/t/1jvcwha 

And part of the Dialog replacement means supporting different ways of getting OUs. The old code returns objects, the new code returns paths.

SferaDev (approved):

PR #224, opened 2022-01-20
[Feature] Replace MultiSelectors with Transfers

SferaDev:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/1p2bgj1

### :memo: Implementation

Created `UserRolesSelector` and `UserGroupsSelector` that show the transfer component inside a `ConfirmationDialog`. Replaced the `Toggle` component for the update strategy to `SegmentedControl` component (added to types) . Also using d2-api and implemented two new use cases: `UpdateUsersRolesUseCase` and  `UpdateUsersGroupsUseCase`.

### :art: Screenshots

image

image


SferaDev:
Thanks @p3rcypj, please check my review and send me a slack if you want to discuss why some of my changes.

@idelcano Please do a thorough functional review and test all possible edge cases of updating user groups and roles.

SferaDev:
> -I cant update or save users roles or groups from MULTIPLE USER, these calls are failing due server error 500 after update, with merge or replace mode.

@p3rcypj Can you review it?

> After change some individually user, if i click on details, the details are pre-loaded from the filter and i didnt see the changes (for example replace all groups).

We are missing a table reload on change, could you review it @p3rcypj 

> possible feature: I have replace the roles from an user using before the search, and only the visible roles was replaced (in my first test 0 was shown, and the user lost all their roles)

I don't fully understand this, could you please elaborate @idelcano

p3rcypj:
I will take look to both errors.

PR #227, opened 2022-02-09
[Feature] Save column order in data store

SferaDev:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/1xwfebr

### :memo: Implementation

Added usecases and methods for: saving selected columns on the data store, and for retrieving them when component renders.


SferaDev:
Thanks @p3rcypj, @idelcano will functionally review the feature

src/webapp/components/user-list-table/UserListTable.tsx
```diff
@@ -222,20 +240,44 @@ export const UserListTable: React.FC<UserListTableProps> = props => {
...
+
+    useEffect(() => {
+        compositionRoot.users.getColumns().run(
+            columns => {
+                setVisibleColumns(columns);
+                onChangeVisibleColumns(columns);
+            },
+            error => snackbar.error(error)
+        );
+    }, [compositionRoot, snackbar, onChangeVisibleColumns]);
```
SferaDev, line 272:
```suggestion
    useEffect(() => {
        return compositionRoot.users.getColumns().run(
            columns => {
                setVisibleColumns(columns);
                onChangeVisibleColumns(columns);
            },
            error => snackbar.error(error)
        );
    }, [compositionRoot, snackbar, onChangeVisibleColumns]);
```

idelcano:
The feature looks great.
But could we add some default columns?
When i open the app the first time looks strange:
screenshot-localhost_8081-2022 02 09-10_47_58


And when add more columns the header looks uggly too:

screenshot-localhost_8081-2022 02 09-10_49_26

SferaDev:
> The feature looks great. But could we add some default columns? When i open the app the first time looks strange: screenshot-localhost_8081-2022 02 09-10_47_58

Indeed, @p3rcypj can you take a look?

> And when add more columns the header looks uggly too:

We will not fix this, happens on all our apps

p3rcypj:
Sure. Should I add the initial columns if there is no column selected?

SferaDev:
> Sure. Should I add the initial columns if there is no column selected?

No, it is more for the user when it enters the first time that the columns will be empty. You can force that by creating a new user and trying to enter the app @p3rcypj

SferaDev:
Thanks @p3rcypj

PR #231, opened 2022-02-17
[Fix] Replace MultiSelectors with Transfers

p3rcypj:
### :pushpin: References

-   **Issue:** Fixes #224 

### :memo: Implementation

Reset loading on error. Fix on removing roles/group.


p3rcypj:
Common roles and groups are set in a constant before the user map so it doesn't have to calculate them on every user.

SferaDev (approved):

PR #232, opened 2022-02-18
Add small improvements to user edit page

nshandra:
### Tasks

- Add accountExpiry back.
- Remove autocomplete from the form.
- Save values from locale fields.


adrianq:
Most of the features have gone into the master branch a long time ago. The only two missing features were:
- Add accountExpiry back.
- Save values from locale fields.
@eperedo have migrated them into https://github.com/EyeSeeTea/user-extended-app/pull/274 so we can safely close this PR

PR #233, opened 2022-06-16
Feature/Implement interface for selecting AND or OR filtering behavior

nshandra:
### :pushpin: References

-   **Issue:** https://app.clickup.com/t/2cbz4q8

### :memo: Implementation

Add a switch button to the "Advanced filters" sub-screen that changes the state of `this.state.rootJunction`.
`this.state.rootJunction === false` means AND filtering and `this.state.rootJunction === true` means OR.

### DIFF
--- a/src/legacy/List/Filters.component.js
+++ b/src/legacy/List/Filters.component.js
@@ -232,12 +232,16 @@ export default class Filters extends React.Component {
                                 />
                             </Grid>
                             <Grid item xs={6} className="control-row switch">
-                                <span>{this.getTranslation("AND")}</span>
-                                <Switch
-                                    onChange={this.setFilter("rootJunction", this.checkboxHandler)}
-                                    checked={rootJunction}
-                                />
-                                <span>{this.getTranslation("OR")}</span>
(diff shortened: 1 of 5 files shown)


src/legacy/List/Filters.component.js
```diff
@@ -213,22 +216,34 @@ export default class Filters extends React.Component {
...
+                                    checked={showOnlyManagedUsers}
+                                />
+                                <Checkbox
+                                    className="control-checkbox"
+                                    label={this.getTranslation("display_only_enabled_users")}
+                                    onCheck={this.setFilter("showOnlyActiveUsers", this.checkboxHandler)}
+                                    checked={showOnlyActiveUsers}
+                                />
+                            </Grid>
+                            <Grid item xs={6} className="control-row switch">
```
jordibc, line 234: I think this is a nice solution! For aesthetic reasons, how about having the Grid items with `xs={8}` and `xs={4}` respectively? (To have the switch button more to the right.)

jordibc (commented):
This is great! I like your solution. You may want to shift a bit the checkboxes, but in any case I'm happy to merge this.

PR #234, opened 2022-06-23
Feature/use filter as and or or

jordibc:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/22k4dbc

### :memo: Implementation

Depending on the direction of the new AND/OR switch gui element, it sets a `rootJunction` property that can be "AND" or "OR", and then it uses it when making the call to `compositionRoot.users.list()`.

### :art: Screenshots

Some filters added in **OR** mode:

Screenshot from 2022-06-23 14-55-33

When changing the same filters to **AND** mode:

Screenshot from 2022-06-23 14-55-42

### DIFF
--- a/src/data/repositories/UserD2ApiRepository.ts
+++ b/src/data/repositories/UserD2ApiRepository.ts
@@ -38,6 +38,7 @@ export class UserD2ApiRepository implements UserRepository {
             filters,
         } = options;
         const otherFilters = _.mapValues(filters, items => (items ? { [items[0]]: items[1] } : undefined));
+        const areFiltersEnabled = _(otherFilters).values().some()
 
         return apiToFuture(
             this.api.models.users.get({
@@ -46,7 +47,7 @@ export class UserD2ApiRepository implements UserRepository {
                 pageSize,
                 query: search !== "" ? search : undefined,
                 filter: otherFilters,
-                rootJunction: filters ? rootJunction : undefined,
+                rootJunction: areFiltersEnabled ? rootJunction : undefined,
                 order: `${sorting.field}:${sorting.order}`,
             })
         ).map(({ objects, pager }) => ({
--- a/src/data/repositories/UserD2ApiRepository.ts
+++ b/src/data/repositories/UserD2ApiRepository.ts
@@ -249,10 +249,10 @@ export class UserD2ApiRepository implements UserRepository {
     private toDomainUser(input: ApiUser): User {
         const { userCredentials, ...user } = input;
         const authorities =
-            _(userCredentials.userRoles?.map(userRole => userRole.authorities))
+            _(userCredentials.userRoles).map(userRole => userRole.authorities)
                 .flatten()
                 .uniq()
-                .value() || [];
+                .value();
 
         return {
             id: user.id,
--- a/src/data/repositories/UserD2ApiRepository.ts
+++ b/src/data/repositories/UserD2ApiRepository.ts
@@ -247,10 +247,11 @@ export class UserD2ApiRepository implements UserRepository {
 
     private toDomainUser(input: ApiUser): User {
         const { userCredentials, ...user } = input;
-        const authorities = _(userCredentials.userRoles?.map(userRole => userRole.authorities))
-            .flatten()
-            .uniq()
-            .value() || [];
+        const authorities =
+            _(userCredentials.userRoles?.map(userRole => userRole.authorities))
+                .flatten()
+                .uniq()
+                .value() || [];
 
         return {
             id: user.id,
--- a/src/data/repositories/UserD2ApiRepository.ts
+++ b/src/data/repositories/UserD2ApiRepository.ts
@@ -38,7 +38,7 @@ export class UserD2ApiRepository implements UserRepository {
             filters,
         } = options;
         const otherFilters = _.mapValues(filters, items => (items ? { [items[0]]: items[1] } : undefined));
-        const areFiltersEnabled = _(otherFilters).values().some()
+        const areFiltersEnabled = _(otherFilters).values().some();
 
         return apiToFuture(
             this.api.models.users.get({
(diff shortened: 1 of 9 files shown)


src/data/repositories/UserD2ApiRepository.ts
```diff
@@ -39,6 +46,7 @@ export class UserD2ApiRepository implements UserRepository {
                 pageSize,
                 query: search !== "" ? search : undefined,
                 filter: otherFilters,
+                rootJunction: filters ? rootJunction : undefined,
```
tokland, line 49: If you go to advanced filters, set mode=AND then mode=OR, the user list shows empty. The problem is here. This line correctly wants to skip `rootJuction` if no filters are enabled (as this breaks OR mode). But `filters` is an object, which in JS is always truthy. Also, even if it had properties, a filter is disabled when it has null/undefined values. Also, it should check `otherFilters`, which is the final value. A proposal:

```
const anyFilterEnabled = _(otherFilters).values().some()
```

src/data/repositories/UserD2ApiRepository.ts
```diff
@@ -39,6 +46,7 @@ export class UserD2ApiRepository implements UserRepository {
...
                 order: `${sorting.field}:${sorting.order}`,
             })
```
tokland, line 51: I guess this problem is know, but I noticed the filter "Display only users I can manage" does nothing. The old code passed the param "canManage=true", but it's not in the new code.

src/data/repositories/UserD2ApiRepository.ts
```diff
@@ -239,10 +247,11 @@ export class UserD2ApiRepository implements UserRepository {
 
     private toDomainUser(input: ApiUser): User {
         const { userCredentials, ...user } = input;
-        const authorities = _(userCredentials.userRoles.map(userRole => userRole.authorities))
-            .flatten()
-            .uniq()
-            .value();
+        const authorities =
+            _(userCredentials.userRoles?.map(userRole => userRole.authorities))
```
tokland, line 251: [minor] Lodash accepts null/undefined values, it considers them empty arrays. So this can be written: `_(userCredentials.userRoles).map(userRole => userRole.authorities)`

src/data/repositories/UserD2ApiRepository.ts
```diff
@@ -239,10 +247,11 @@ export class UserD2ApiRepository implements UserRepository {
...
+                .flatten()
+                .uniq()
+                .value() || [];
```
tokland, line 254: For the same reason, this `|| []` is not necessary. TS can attest to that.

src/legacy/List/Filters.component.js
```diff
@@ -231,16 +233,19 @@ export default class Filters extends React.Component {
...
-                                        onChange={this.setFilter("rootJunction", this.checkboxHandler)}
-                                        checked={rootJunction}
+                                        onChange={() => {
+                                            const newRootJunction = rootJunction === "AND" ? "OR" : "AND";
+                                            this.setState({ rootJunction: newRootJunction }, this.notifyParent);
+                                        }}
+                                        checked={rootJunction === "AND"}
                                     />
-                                    <span>{this.getTranslation("OR")}</span>
+                                    <span>{this.getTranslation("AND")}</span>
```
tokland, line 248: I am not familiar with the infrastructure of this project, but I see that we have a `public/old-i18n` and the gettext infrastructure is there. Maybe we should be using then `i18n.t` on new code?

src/data/repositories/UserD2ApiRepository.ts
```diff
@@ -39,6 +46,7 @@ export class UserD2ApiRepository implements UserRepository {
                 pageSize,
                 query: search !== "" ? search : undefined,
```
tokland, line 47: The value of `rootJunction` does not seem to affect how the query+filter intersection is being applied (it's always an AND). 

So if we really want to support OR when both kind of filters are enabled, we'll need to do some programmatic filtering. That's not trivial, as now the `pager` from the API is wrong, and AFICS we don't have any way to calculate it other than: getting all the users+ apply programmatic filter + build the pager.

jordibc: My understanding is that we want the AND/OR to be applied only to the advanced filters, not to the query + advanced filters intersection. But I'm not sure about it, the request came via @ifoche (and I can't check the original request mentioned in https://app.clickup.com/t/22k4dbc), so I'll ask him.

src/data/repositories/UserD2ApiRepository.ts
jordibc: That's very interesting, I certainly didn't know it, and I was a bit puzzled about what that filter meant (and still am).

I checked and it seems that in `UserD2ApiRepository.ts` there has never been a `canManage` (nor `showOnlyManagedUsers` or anything like that). Maybe the bug came when this part was refactored to follow a clean architecture?

In any case, I think that should be filled as a (separate) bug/task.

src/legacy/List/Filters.component.js
jordibc: I wasn't familiar with the distinction of `old-i18n` vs `i18n.t`. The rest of the component uses `this.getTranslation`, and I'm happy to see how to change it, but I'm not sure this PR would be the place? What would you prefer?

jordibc (commented):
Thanks @tokland !

The bug was exactly as you described it and I fixed it with your solution. Also, it was nice to know about the simplification coming from lodash semantics.

src/legacy/List/Filters.component.js
tokland: I see, we can leave this.getTranslation for existing components and i18n.t for the new ones.

tokland (approved):
All good! Yeah, @ifoche will decide about the and/user with the search text.

ifoche:
Hi @jordibc @tokland . Re the AND/OR... I think a normal user would understand that the AND/OR logic would apply to everything in the filters. I understand that there's a technical issue that prevent from doing it. But if we go ahead with this solution, then we should explicitly inform the user that AND/OR logic is not applied to the non-advanced filter, which is always AND. So my answer is... if we can propagate the AND/OR without risking a big performance issue, then please go ahead with that. If not, then please try to find a way to clearly inform the user of that situation in the front-end

ifoche (approved):
thanks @jordibc @tokland

PR #235, opened 2022-06-24
Feature/add export import json

MarieJousse:
### :pushpin: Issue
Task https://app.clickup.com/t/2f1wztt

### :memo: Implementation
Add an option to export the users selected on a JSON file and not only on a CSV file. 
It gathers all the users according to the filters that are applied and uses JSON.stringify() to respect Json structure. 
Finally it saves the file. 

The import option now accepts both CSV and JSON files.

### :art: Screenshots
image


jordibc:
Thanks @nshandra !

I found a few issues:

* Importing a json that I have exported (and optionally modified), does not work (see screenshot)

Screenshot from 2022-07-13 12-05-05

* The arrows for import/export have the wrong direction

This seems to be because in `src/legacy/components/ImportExport.component.js` the imports take the wrong names. It should be:

```ts
import ImportIcon from "material-ui/svg-icons/navigation/arrow-upward";
import ExportIcon from "material-ui/svg-icons/navigation/arrow-downward";
```

(among other few changes)

Finally, There seems to be a few functions that "want to" go together:

* namesFromCollection - namesArrayFromCollection
* getPlainJsonUser - getPlainUser
* exportToCsv - exportToJson

I have an idea of how each pair can be essentially the same function (used slightly different when called). I'm happy to elaborate more if you want, unless you see it already and wanna give it a try!

src/legacy/components/ImportExport.component.js
```diff
@@ -11,7 +11,7 @@ import FileSaver from "file-saver";
 import moment from "moment";
```
jordibc, line 11: Please see the comments in https://github.com/EyeSeeTea/user-extended-app/pull/235#issuecomment-1188381596 for the review. Thanks!

ifoche (changes requested):
@nshandra could you please review this PR following @jordibc comments? Thanks

src/legacy/components/ImportExport.component.js
nshandra: Fixed importFromJson TypeError and icons.
Merged functions: namesFromCollection & namesArrayFromCollection => namesFromCollection, getPlainJsonUser & getPlainUser => getPlainUser and exportToCsv & exportToJson => exportUsers.
Add csvWarningsToJsonWarning function to adapt the CSV warnings to JSON warnings.

csv warnings format:
`Value not found: ${missingValue} [username=${username || "-"} csv-row=${rowIndex} csv-column=${field}]`

json warnings format:
`Value not found: ${missingValue} [username=${username || "-"} json-key=${field}]`

ifoche (approved):
thanks @nshandra

PR #236, opened 2022-06-30
[User-Extended] Apply filters to the export function

nshandra:
### :pushpin: References

-   **Issue:** https://app.clickup.com/t/2f1tj5j

### :memo: Implementation

The `src/legacy/components/ImportExport.component.js` property `filterOptions` was missing the filter by name query, that's why the filter was not applied to the export.

### DIFF
--- a/src/legacy/List/List.component.js
+++ b/src/legacy/List/List.component.js
@@ -206,9 +206,9 @@ export class ListHybrid extends React.Component {
 
     filterList = () => {
         const order = this.state.sorting ? this.state.sorting[0] + ":i" + this.state.sorting[1] : null;
-        const { filters } = this.state;
+        const { filters, query } = this.state;
 
-        this.setState({ isLoading: true, listFilterOptions: { order: order, ...filters } });
+        this.setState({ isLoading: true, listFilterOptions: { order: order, query: query, ...filters } });
     };
 
     convertObjsToMenuItems = objs => {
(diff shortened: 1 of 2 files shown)


jordibc:
Thanks @nshandra !

This looks great. I didn't even know there were those functionalities in `ImportExport` (and `exportToCsv`) and it took me a while to understand how that was working. These changes use it very nicely.

A very minor comment, I might call `_updateSearch` something like `_updateQuery`, since `query` is the variable updated, and it would be more consistent with the naming of other similar functions in that file. But I'm not opposed to its current name.

I also tested the thoroughly the functionality and it works perfectly (in case @ifoche wants to skip checking).

src/legacy/List/List.component.js
```diff
@@ -268,6 +268,10 @@ export class ListHybrid extends React.Component {
         this.setState({ visibleColumns });
     };
 
+    _updateSearch = query => {
```
jordibc, line 271: A very minor comment, I might call `_updateSearch` something like `_updateQuery`, since `query` is the variable updated, and it would be more consistent with the naming of other similar functions in that file. But I'm not opposed to its current name.

jordibc (approved):
Thanks @nshandra !

This looks great. I didn't even know there were those functionalities in `ImportExport` (and `exportToCsv`) and it took me a while to understand how that was working. These changes use it very nicely.

Other than the very minor comment about `_updateSearch` / `_updateQuery`, there's nothing I can add.

I also tested the thoroughly the functionality and it works perfectly (in case @ifoche wants to skip checking).

src/legacy/List/List.component.js
nshandra: Yes, it makes sense. Done.

ifoche (approved):
thanks @nshandra @jordibc

PR #237, opened 2022-07-07
Bugfix/broken group and roles dialog for import

nshandra:
### :pushpin: References

-   **Issue:** [[User-Extended] fix groups and roles popup on intermediate screen for import](https://app.clickup.com/t/2jm5d42)

### :memo: Implementation

The code for _userGroups_ and _userRoles_ was removed, i guess in order to migrate to the newer components.
Given that this [issue](https://app.clickup.com/t/2f1wzrc) will update the import dialog, i will reuse the old code for the fix.

Also some missing translations were added.

### :fire: Testing

No new code was added, only re-added old code, only functional testing should be needed.

### DIFF
--- a/public/old-i18n/i18n_module_ar.properties
+++ b/public/old-i18n/i18n_module_ar.properties
@@ -1,5 +1,6 @@
 actions=Actions
 assign_all=Assign all
+remove_all=Remove all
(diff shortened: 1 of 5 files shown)


ifoche (approved):
thanks @nshandra

PR #239, opened 2022-07-13
Feature/ou tree swith to shortnames

jordibc:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/2ghev8r

This just points to the new version of d2-ui-components with the updated OrgUnitsSelector that has the new property to show the corresponding switch to use short names.

For the implementation details, screenshots and all, see https://github.com/EyeSeeTea/d2-ui-components/pull/227

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
         "@dhis2/d2-i18n-generate": "1.2.0",
         "@dhis2/ui": "7.8.1",
         "@eyeseetea/d2-api": "1.11.0",
-        "@eyeseetea/d2-ui-components": "2.6.12-beta.3",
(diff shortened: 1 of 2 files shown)


ifoche (changes requested):
@jordibc I'm getting an error here when compiling... coud you have a look? Thanks

ifoche (approved):
thanks @jordibc and @tokland . It worked like a charm

@jordibc just so we don't forget about it. I will create issues for applying the same logic to all components in the app that work with OUs. At this moment it is only working like that in the filters, but e.g. in the OU assignment, the slider doesn't appear.

PR #240, opened 2022-07-21
Bugfix/open id export import not working

nshandra:
### :pushpin: References

Task https://app.clickup.com/t/2k5g5xm

### :memo: Implementation
OpenID was missing from the query parameters for export, and didn't figured as a column or `userCredentialsFields` value for import.

For now the OpenID field must be unique for each user on import, i didn't managed to find out if this is intended or if its a quirk of the code. DHIS2 Users allows to repeat OpenIDs, but on the other hand, it looks like OpenID should map one to one to users in order to identify them.
Links: 
[OpenID Connect (OIDC) configuration](https://docs.dhis2.org/en/manage/performing-system-administration/dhis-core-version-236/installation.html#install_oidc_configuration)
[Configure OpenID Connect with Okta](https://docs.dhis2.org/en/topics/tutorials/configure-oidc-with-okta.html)

### DIFF
--- a/public/old-i18n/i18n_module_ar.properties
+++ b/public/old-i18n/i18n_module_ar.properties
@@ -135,3 +135,4 @@ select_at_least_one_user=Select at least one destination user
 AND=AND
 OR=OR
 Filtering_behavior=Filtering behavior 
+open_id=Open ID
--- a/public/old-i18n/i18n_module_en.properties
+++ b/public/old-i18n/i18n_module_en.properties
@@ -135,3 +135,4 @@ select_at_least_one_user=Select at least one destination user
 AND=AND
 OR=OR
 Filtering_behavior=Filtering behavior 
+open_id=Open ID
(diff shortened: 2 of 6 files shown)


jordibc (commented):
I'm editing this comment in case you've seen it already. As I say in my next one, it is working well, so I'm closing this PR. Thanks!

---

Thanks @nshandra !

The changes in the code look good, but I'm having a problem when trying it. This is what I did:

* Download an instance of dhis2 2.36 (`docker.eyeseetea.com/widpit/dhis2-data:2.36-WIDP-DEV` in my case).
* Exported users to a csv that includes openId
* Modified the csv to put a single new user
* Imported the csv

I'm attaching the csv that I use to import:

[test_openid.csv](https://github.com/EyeSeeTea/user-extended-app/files/9191796/test_openid.csv)

The import is successful, but the openId field is empty:

Screenshot from 2022-07-26 18-45-21

I also found problems when running it against dhis2 2.38, but they seem unrelated to your changes.

How do you test it? Can you give me an example of container and csv file?

jordibc:
Sorry, it does work (I had been trying to figure out what was happening with 2.38 for a while, and at one point I started testing the development branch, and when I did the final test I forgot to update to this branch!).

jordibc (approved):
All is fine.

PR #241, opened 2022-09-29
Use switch to show short names in all uses of the OrgUnitsSelector component

jordibc:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/2ghev8r

### :memo: Implementation

Trivial. Just added `showNameSetting={true}` to the OrgUnitsSelector component in the OrgUnitDialog component and the OrgUnitSelectorFF component.

### :art: Screenshots

When clicking on "Assign to organisation units":

Screenshot from 2022-09-29 11-32-01

When clicking on "Assign to data view organisation units":

Screenshot from 2022-09-29 11-32-16

When editing a user:

Screenshot from 2022-09-29 11-46-46

### DIFF
--- a/src/legacy/List/organisation-unit-dialog/OrgUnitDialog.component.js
+++ b/src/legacy/List/organisation-unit-dialog/OrgUnitDialog.component.js
@@ -124,6 +124,7 @@ class OrgUnitDialog extends React.Component {
                         filterByProgram: false,
                         selectAll: false,
                     }}
+                    showNameSetting={true}
                 />
             </ConfirmationDialog>
         );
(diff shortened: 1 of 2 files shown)


ifoche (changes requested):
@jordibc there's one place where this is not yet implemented. This is when you export in csv and then import that csv back. In that intermediate dialog that you get when you are importing, if you tried to import the OUs of the users, you can click in the field and a OU tree appears...
screenshots_1665072219_1803_06102022_1227x496
screenshots_1665072211_1803_06102022_1254x747
As you see, in that OU tree there's no selection of the shortname/longname

ifoche (approved):
following a conversation with @jordibc , the reason why the import from CSV cannot get the new OU shortname/longname selector is because it's using a legacy component and it would take long to replace it by the new one. I'm then approving this PR and leaving for a future task the refactor of it

PR #242, opened 2022-10-06
Comment out all uses of accountExpiry.

jordibc:
This way, when we update a user we do not send an empty value for accountExpiry.

Dhis2 interprets now an empty value as setting it to its first date, 1970-01-01, and that stops the users from being able to connect (since the date is in the past).

### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/3jv4b34

### :memo: Implementation

Commented out all appearances of `accountExpiry`.

### :art: Screenshots

In the POST request when modifying a user, there is no `accountExpiry` anymore (and thus dhis2 doesn't set it to 1970).

image


ifoche (approved):
thanks @jordibc . Worked like a charm

PR #247, opened 2023-01-23
Fix/The copy in users fails in 2.38

nshandra:
### :pushpin: References

-   **Issue:** Closes [[User-Extended App] the copy in users fails in 2.38](https://app.clickup.com/t/3dehbbg)

### :memo: Implementation
- With DHIS 2.38 the users retrieved in `getUserInfo` were missing the `userCredentials` field, making the "copy in user" operation fail, this happens because the default fields of `api/users` no longer include the aforementioned `userCredentials` field.
- Fixed replicate in table/template actions, witch suffered the same issue about missing `userCredentials`. Removed replicating the following fields: `externalAuth`, `openId`, `ldapId`, `lastLogin` , `created` and `createdBy` as they introduced false information and made the replication fail in the case of the external auth fields.
- Fixed Delete action.
- Fixed Disable action.
- Tested that the changes don't break the app for DHIS 2.36.


nshandra:
Reverted to Draft to add additional fixes.

ifoche (changes requested):
@nshandra seems like the import with CSV is broken. Could you please give it a look?

To replicate it, you can use the following in your .env.local
```
REACT_APP_DHIS2_BASE_URL=https://play.dhis2.org/2.38.2.1
REACT_APP_DHIS2_AUTH='admin:district'
```
and use the following csv file to be imported:
[users-2023-02-13_15-47-03.csv](https://github.com/EyeSeeTea/user-extended-app/files/10723340/users-2023-02-13_15-47-03.csv)

The file contains 2 existing users and 1 new. The 2 existing have 1 property modified each, and you would have to activate the "Overwrite existing users" flag if you want to be able to proceed, and the new one will need to be added with a password. When you click on "Import" you get this error in the console:
screenshots_1676300029_1553_13022023_513x150

ifoche (changes requested):
@nshandra seems like the copy in user for user roles and groups is not properly working. Could you please have a look to it?

nshandra:
The CopyInUserDialog component was set up to handle POST errors as exceptions but it didn't raise any when that happens.
I will port this error message to the same format as the error message for "replicate users from *" in another PR, as we need a working 2.38 version ASAP.

ifoche (approved):
thanks @nshandra

PR #249, opened 2023-02-22
Fix empty userRoles POST when using the edit page

nshandra:
### :pushpin: References

-   **Issue:** [Bug on user-Extended app for user creation/modification](https://app.clickup.com/t/860q01d6x)

### :memo: Implementation
- Fix user edit not working in 2.38 (still broken in 2.36, gives error 500)
- Fix import missing some fields
- Disable new user button as this function doesn't work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "user-extended-app",
     "description": "DHIS2 Extended User app",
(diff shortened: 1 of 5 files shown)


ifoche (approved):
thanks @nshandra

PR #257, opened 2023-06-29
Validate if user has roles, groups or OUS in export process

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/860r74hyd

### :memo: Implementation

Original bug about importing passwords cannot be replicated, but we found a bug related to the export process (which is a common previous step for import) so this PR includes the fix.


ifoche (approved):
thanks @eperedo

PR #270, opened 2024-01-02
add createdBy and lastUpdatedBy columns

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8693e3wam

Columns in table

image

Columns in selector

image


src/data/repositories/UserD2ApiRepository.ts
```diff
@@ -159,6 +166,18 @@ export class UserD2ApiRepository implements UserRepository {
         });
     }
 
+    private getUserByIds(userIds: Id[]) {
```
tokland, line 169: getUsersById

tokland (approved):
All good! (just a minor proposal of rename)

MiquelAdell:
@eperedo just so I know, I've seen this code change:
```// in version 2.38 stats and typeReports are inside a response object
type Dhis2Response = MetadataResponse & {
    response?: { stats: MetadataResponse["stats"]; typeReports: MetadataResponse["typeReports"] };
};
```
Will that mean that the app will not be compatible with 2.37? or are we doing some kind of check to get stats from one part or the other?

I'm ok either way but I need to know what versions are we supporting once I release the app

eperedo:
> @eperedo just so I know, I've seen this code change:
> 
> ```
> type Dhis2Response = MetadataResponse & {
>     response?: { stats: MetadataResponse["stats"]; typeReports: MetadataResponse["typeReports"] };
> };
> ```
> 
> Will that mean that the app will not be compatible with 2.37? or are we doing some kind of check to get stats from one part or the other?
> 
> I'm ok either way but I need to know what versions are we supporting once I release the app

We're checking the response so we can support both versions.

PR #272, opened 2024-02-05
Show simpler 'replicate user' error message

davidbgomes:
### :pushpin: References

-   **Issue:** Closes [[User-Extended App] - Errors when replicating users are not caught and shown, but instead, a big json file is thrown](https://app.clickup.com/t/8693ncwng)

### :memo: Implementation

- Use snackbar to show a simpler error message.

image

### DIFF
--- a/src/legacy/components/ReplicateUserFromTable.component.js
+++ b/src/legacy/components/ReplicateUserFromTable.component.js
@@ -41,7 +41,12 @@ class ReplicateUserFromTable extends React.Component {
             snackActions.show({ message });
             return null;
         } else {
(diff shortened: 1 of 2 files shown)


eperedo (changes requested):
Thanks @davidbgomes, code looks good. But I'm still getting this big json alert from the `replicate from user` option, not sure if this change was something specific for the `replicate from table` option?

image

davidbgomes:
Thanks @eperedo, you're right it is for both replications. Should be good now!

eperedo (approved):

PR #281, opened 2024-04-29
Add some basic tests

p3rcypj:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8694ekvck

### :memo: Implementation

Added some basic tests for `User` entity and `ListUsersUseCase`

### :fire: Testing

Docker used: `docker.eyeseetea.com/eyeseetea/dhis2-data:2.37.9-play`


p3rcypj:
I've used Test Doubles for the `useCase` manually. A bunch of "tasks" that come to my mind would be:
- [ ] Finish `ListUsersUseCase.spec.ts` with the remaining parameters the `options` argument can be used with
- [ ] Make Test Doubles being created manually to be created with `ts-mockito`
- [ ] Choose another simple `useCase` to add tests to it
- [ ] Create tests for the UI components

xurxodev (changes requested):
Thanls @p3rcypj 

The idea in general is good but in test for ListUserUseCase there are some issues to fix:

- I would avoid the use of `as UserRepository` in givenAListOfUsers, givenAListOfUsersWithPageSizeOption, givenAPager and givenAListOfUsersBySearchOption because is unsafe. I understand that the goal is not implement unnecessary methods but if list function change in the future in the repository, the test fail to execute but not in compilation time.

A better solution for me is to create a specific class for all tests named UserStubRepository implementing UserRepository. Leave empty the non used functions and that receive by the constructor the expected objects and pager for every test.

MiquelAdell:
superseeded by #283, #288 and #292

PR #288, opened 2024-05-30
chore: add test for `SaveUserOrgUnitUseCase`

delemaf:
### :pushpin: References

-   **Issue:** https://app.clickup.com/t/8694ekvck
- 8694ekvck

### :memo: Implementation

- `ts-mockito` installed
- test on `SaveUserOrgUnitUseCase`

### :fire: Testing

To launch the specific test `yarn test SaveUserOrgUnitUseCase.spec.ts`

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
         "rxjs": "6.6.7",
         "styled-components": "5.3.3",
         "styled-jsx": "4.0.1",
+        "ts-mockito": "^2.6.1",
         "use-deep-compare-effect": "1.8.1"
     },
     "devDependencies": {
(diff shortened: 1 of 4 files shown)


delemaf:
I didn't understand why we overwrite org unit `name` and `path`, is it a normal behaviour? @eperedo 

Example:
I wanted to replace user's org units with this org unit : 
`{
        id: "ImspTQPwCqd",
        name: "Sierra Leone",
        path: ["ImspTQPwCqd"],
 }`
When I look what's saved in users I've :
`{
        id: "ImspTQPwCqd",
        name: "",
        path: [],
 }`

eperedo (approved):
Looks good, thanks @delemaf!

eperedo:
@delemaf that was because the org. unit selector component only returns the id of the selected orgunits and to avoid making an extra request for orgunits we're just sending the ids. This is not a bug because if you want to save the orgunits for a user DHIS only needs the ids: 

image

It could be an improvement to get the full information of the org. units but not necessary for this particular feature.

delemaf:
> @delemaf that was because the org. unit selector component only returns the id of the selected orgunits and to avoid making an extra request for orgunits we're just sending the ids. This is not a bug because if you want to save the orgunits for a user DHIS only needs the ids:
> 
> image
> 
> It could be an improvement to get the full information of the org. units but not necessary for this particular feature.

Alright I understand, I just wanted to be sure it doesn't erase org units name 👌

PR #289, opened 2024-05-31
feature: export migration

delemaf:
### :pushpin: References

-   **Issue:** https://app.clickup.com/t/8694mneh7
- 8694mneh7

### :memo: Implementation
⚠️ This PR only concern **export**, the **import** will be done in a separate PR (it'll be to big)
- new `ImportExport.tsx` component (old `ImportExport.component.js` is deleted)
- new `ExportUsersUseCase` that migrate related legacy function in `userHelpers.js`
- tests for use case
- custom hook `useExportUsers`
- When orgUnitsField (ex: code) doesn't exist it fallback to `name` field (I encounter a org unit without a `code` and it export `null` so I think it could be a workaround to use name)

https://github.com/EyeSeeTea/user-extended-app/assets/6376073/82c67b2b-a292-4e09-be8e-3747230b3a48

### :fire: Testing
- Try to export in CSV and JSON
- Filter with name for example and try to export again, you should only have the filtered users
- Import should still work
- `yarn test` should still pass


src/webapp/hooks/userHooks.ts
```diff
@@ -69,3 +78,38 @@ export function useSaveUsersOrgUnits(props: UseSaveUsersOrgUnitsProps) {
...
+
+export const useExportUsers = (props: UseExportUsersProps) => {
+    const { onSuccess, columns, filterOptions, orgUnitsField } = props;
+
+    const { compositionRoot } = useAppContext();
+    const snackbar = useSnackbar();
+    const loading = useLoading();
+
+    const exportUsers = React.useCallback(
+        (name: string, format: AllowedExportFormat, isEmptyTemplate = false) => {
```
eperedo, line 90: Let's be explicit about the type to help typescript infer the type (see isEmptyTemplate: any in the image)

image

`(name: string, format: AllowedExportFormat, isEmptyTemplate: boolean)`

src/domain/usecases/ExportUsersUseCase.ts
```diff
@@ -0,0 +1,154 @@
...
+
+export type AllowedExportFormat = "json" | "csv";
+
+export type ExportUsersUseCaseOptions = {
+    name: string;
+    columns: ColumnMappingKeys[];
+    filterOptions?: ListOptions;
+    format: AllowedExportFormat;
+    orgUnitsField: OrgUnitKey;
+    isEmptyTemplate?: boolean;
```
eperedo, line 153: Let's try to avoid optional types. `isEmptyTemplate: boolean`.

src/webapp/components/import-export/ImportExport.tsx
```diff
@@ -0,0 +1,109 @@
...
+            });
+    };
+
+    return (
+        <div className="data-table-import-export">
+            <IconButton onClick={openMenu} tooltipPosition="bottom-left" tooltip={i18n.t("Import/Export")}>
+                <ImportExportIcon />
+            </IconButton>
+
+            {isProcessing && <ModalLoadingMask />}
```
eperedo, line 67: Normally we use the `useLoading` hook for displaying the loading screen.

src/webapp/components/import-export/ImportExport.tsx
```diff
@@ -0,0 +1,109 @@
...
+};
+
+export type FilterOption = { search: string; sorting: { field: string; order: "asc" | "desc" } };
+
+export type ImportExportProps = {
+    columns: ColumnMappingKeys[];
+    filterOptions: FilterOption;
+    onImport: (result: any) => void;
+    maxUsers: number;
+    settings: Record<string, any>;
```
eperedo, line 108: I think we can use the Settings class as a type, right? `src/legacy/models/settings.js`

```ts
class Settings {
 ...
}

export type ImportExportProps = {
    columns: ColumnMappingKeys[];
    filterOptions: FilterOption;
    onImport: (result: any) => void;
    maxUsers: number;
    settings: Settings;
};
```

package.json
```diff
@@ -66,10 +66,12 @@
         "@testing-library/cypress": "8.0.2",
         "@testing-library/jest-dom": "5.15.1",
         "@testing-library/react": "12.1.2",
+        "@types/file-saver": "^2.0.7",
```
eperedo, line 69: We have versions: `2.0.5` and `5.3.1` for fileSaver and paparse, but different versions for the @types: 2.0.7 and 5.3.14. Can we match both versions? If there's a problem to do it at least can we use exact versions for the @types? (removing the caret prefix)

src/domain/usecases/ExportUsersUseCase.ts
```diff
@@ -0,0 +1,154 @@
...
+    dataViewOrganisationUnits: i18n.t("OUOutput"),
+    searchOrganisationsUnits: i18n.t("OUSearch"),
+    disabled: i18n.t("Disabled"),
+    status: i18n.t("Status"),
+    openId: i18n.t("Open ID"),
+    createdBy: i18n.t("Created by"),
+    lastModifiedBy: i18n.t("Last modified by"),
+};
+
+export class ExportUsersUseCase implements UseCase {
```
eperedo, line 39: Let's remove the `implements UseCase` since most of the time this class is only going to have the execute method.

eperedo (changes requested):

eperedo:
Also I'm getting this error after running yarn test:

```
ExportUsersUseCase › should return a blob and filename when exporting users with JSON format
expect(received).toEqual(expected) // deep equality

    - Expected  - 3
    + Received  + 3

    @@ -16,17 +16,17 @@
                  "OU_525"
              ],
              "dataViewOrganisationUnits": [
                  "OU_525"
              ],
    -         "created": "2013-03-11 10:51:41",
    +         "created": "2013-03-11 04:51:41",
    -         "lastUpdated": "2024-06-04 08:25:46",
    +         "lastUpdated": "2024-06-04 01:25:46",
              "apiUrl": "/dhis2/api/users/oXD88WWSQpR.json",
              "searchOrganisationsUnits": [
                  "Approved School CHP"
              ],
    -         "lastLogin": "2013-12-30 10:16:43",
    +         "lastLogin": "2013-12-30 04:16:43",
              "status": "Active",
              "disabled": false,
              "lastModifiedBy": "John Traore"
          }
      ]
```

src/data/repositories/UserD2ApiRepository.ts
```diff
@@ -199,6 +199,40 @@ export class UserD2ApiRepository implements UserRepository {
...
+                page,
+                pageSize,
+                paging: false,
+                query: search !== "" ? search : undefined,
+                canManage: canManage === "true" ? "true" : undefined,
+                filter: otherFilters,
+                rootJunction: areFiltersEnabled ? rootJunction : undefined,
+                order: `${sortingField}:${sorting.order}`,
+            })
+        ).map(({ objects }) => objects.map(user => this.toDomainUser(user)));
```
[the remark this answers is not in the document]
delemaf: I'm merged `developpement` to fix this

src/domain/usecases/ExportUsersUseCase.ts
delemaf: I initially tried to specified `isEmptyTemplate: boolean = false` but ESlint fixed it in `isEmptyTemplate = false`, like it was explicit. I've update it with default value inside the function instead in params

src/domain/usecases/ExportUsersUseCase.ts
```diff
@@ -0,0 +1,153 @@
...
+
+export type AllowedExportFormat = "json" | "csv";
+
+export type ExportUsersUseCaseOptions = {
+    name: string;
+    columns: ColumnMappingKeys[];
+    filterOptions?: ListOptions;
+    format: AllowedExportFormat;
+    orgUnitsField: OrgUnitKey;
+    isEmptyTemplate?: boolean;
```
eperedo, line 152: For this case what I mean is to remove the optional parameter so it should be `isEmptyTemplate: boolean` (without the `?`). We try to avoid this kind of parameters because is easy to forget sending the value.

Example:

```ts
function test(options: OptionsType) {
    console.log(options.isEmpty)
}
```
```ts
type OptionsType = { name: string; isEmpty: boolean; }

test({ name: "", isEmpty: true }) // pass
test({ name: "", })  // error
```

with optional parameter

```ts
test({ name: "", isEmpty: true }) // pass
test({ name: "", })  // pass. 
// Not a big deal for a boolean since undefined/false evaluate to false, 
// but as a convention we try to avoid it for every case.
```

Does it make sense?

src/webapp/hooks/userHooks.ts
```diff
@@ -122,3 +132,45 @@ export function useCopyInUser(props: UseCopyInUserProps) {
...
+
+export const useExportUsers = (props: UseExportUsersProps) => {
+    const { onSuccess, columns, filterOptions, orgUnitsField } = props;
+
+    const { compositionRoot } = useAppContext();
+    const snackbar = useSnackbar();
+    const loading = useLoading();
+
+    const exportUsers = React.useCallback(
+        (name: string, format: AllowedExportFormat, isEmptyTemplate?: boolean) => {
```
eperedo, line 144: Same here: `isEmptyTemplate: boolean`

eperedo (changes requested):
Just a minor thing about optional type.

src/domain/usecases/ExportUsersUseCase.ts
delemaf: Yes I understand it maybe be more explicit 👍

src/domain/usecases/__tests__/ExportUsersUseCase.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+        exportUsersUseCase = new ExportUsersUseCase(instance(userRepositoryMock));
+    });
+
+    it("should return an empty template when isEmptyTemplate is true", async () => {
+        const options: ExportUsersUseCaseOptions = {
+            name: "empty-template",
+            columns: columnsAvailableToExport,
+            format: "csv",
+            orgUnitsField: "code",
+            isEmptyTemplate: true,
```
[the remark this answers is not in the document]
delemaf: I didn't saw that, thanks!

delemaf:
Everything is fixed 🚀

PR #287, opened 2024-05-29
feature: copy in user feature migration

delemaf:
### :pushpin: References

-   **Issue:** https://app.clickup.com/t/8694mneh7
- 8694mneh7

### :memo: Implementation
Copy in user feature migration from javascript to typescript:
Old implementation was in component `CopyInUserDialog.js`
New implementation :
- New `CopyInUserDialog.tsx` component
- `UserListTable.tsx` updated 
- New `CopyInUserUseCase.ts` with a unit test
- Added `UserRepository.listAll`

### :video_camera: Screenshots/Screen capture
https://github.com/EyeSeeTea/user-extended-app/assets/6376073/cd8fe648-b144-4bc7-b055-eab5975eaab3

### :fire: Testing
Tested with `docker.eyeseetea.com/eyeseetea/dhis2-data:2.38-widp-hwf-dq --core-image=docker.eyeseetea.com/eyeseetea/dhis2-core:2.38.5`

You can run test for `CopyInUserUseCase` with `yarn test CopyInUserUseCase.spec.ts`

PR #293, opened 2024-06-06
feature: migrate copy in users UI users selector

delemaf:
### :pushpin: References

-   **Issue:** https://app.clickup.com/t/8694mneh7
- 8694mneh7

### :memo: Implementation
⚠️ Merge #287 before
Migrate multiple users selector UI component following this proposition :
https://github.com/EyeSeeTea/user-extended-app/pull/287#issuecomment-2139088849

https://github.com/EyeSeeTea/user-extended-app/assets/6376073/c13f13fe-a614-4f2a-ad7b-b08119e78296

### :fire: Testing
Same test than #287


eperedo (changes requested):
Thanks, @delemaf. It looks good, but I think there was a confusion about the library. Our library is [d2-ui-components](https://github.com/EyeSeeTea/d2-ui-components?tab=readme-ov-file#multiple-selector), and we have a `<MultipleSelector />` component.

Do you think you can take a look at it and, if possible, use that component instead of the one from d2-ui? You can see an [example in bulk load](https://github.com/EyeSeeTea/Bulk-Load/blob/dd92964484c77a9a0ddef5f268183fd02f7136c1/src/webapp/components/settings/DataFormTemplateAssignDialog.tsx#L104). Let me know if you have any questions.

src/webapp/components/copy-in-user-dialog/CopyInUserDialog.tsx
```diff
@@ -79,53 +64,65 @@ export const CopyInUserDialog: React.FC<CopyInUserDialogProps> = props => {
...
-                <MultiSelector
-                    d2={d2}
-                    height={300}
-                    onChange={setSelectedUsersIds}
-                    options={getOptions()}
-                    ordered={false}
-                    searchFilterLabel={i18n.t("Search by name")}
-                    classes={{
-                        searchField: useStyles().searchFieldOverride,
-                    }}
```
delemaf, line 100: > Thanks, @delemaf. It looks good, but I think there was a confusion about the library. Our library is [d2-ui-components](https://github.com/EyeSeeTea/d2-ui-components?tab=readme-ov-file#multiple-selector), and we have a `<MultipleSelector />` component.
> 
> Do you think you can take a look at it and, if possible, use that component instead of the one from d2-ui? You can see an [example in bulk load](https://github.com/EyeSeeTea/Bulk-Load/blob/dd92964484c77a9a0ddef5f268183fd02f7136c1/src/webapp/components/settings/DataFormTemplateAssignDialog.tsx#L104). Let me know if you have any questions.

It was initially a `MultiSelector` from d2-ui-components but I saw a more recent code with d2-ui that looked cleaner, that's why I asked question to @MiquelAdell [here](https://github.com/EyeSeeTea/user-extended-app/pull/287#issuecomment-2139088849)

eperedo: Got it. My bad I thought your initial question was for migrate from `dhis-ui` to `d2-ui-components` (our library). I'm not sure about this. I think is safe to keep using our library since we have control to do any modification, but I will leave this decision up to @MiquelAdell.

eperedo (approved):

PR #294, opened 2024-06-18
include timestamp to url to avoid cache

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8694vdgyw

### :memo: Implementation

- [x] add a "v" (short for version) parameter to the users endpoint to avoid dhis2 cache. The value will be a timestamp.

image

### :fire: Testing

Select any user and edit roles from the `Edit page` or the `Assign roles` action from the table.

#8694vdgyw


ifoche:
Task linked: [CU-8694vdgyw Update user before Save with non-cached response from DHIS2](https://app.clickup.com/t/8694vdgyw)

Ramon-Jimenez (approved):
Thx @eperedo

PR #295, opened 2024-06-25
Feature: import migration

delemaf:
### :pushpin: References

-   **Issue:** https://app.clickup.com/t/8694mneh7 #8694mneh7 

### :memo: Implementation
- `ImportTable` have been migrated almost everything have been reimplemented with final-form (instead of old custom form builder)
- `ImportUsersUseCase` now build a safe merge of db users and imported users (using save from user repository)

#### ⚠️ There's few things that are out of my knowledge
- When override switch is clicked I'm not able to trigger validation so username is not in error state
- What userHelpers process, `importFromCsv` I started to migrate but couldn't make it work as it should (maybe we could migrate it in another iteration)
- When importing file, dialog open blank before useGetAllUsers load, I use loading.show() but didn't manage to make it work
- On import user doesn't seems to update but same thing happen in bulk user update so it's maybe my instance

https://github.com/EyeSeeTea/user-extended-app/assets/6376073/dced5437-9daa-47f3-93ae-353e6c97e7f7


ifoche:
Task linked: [CU-8694mneh7 Improvements to user extended to avoid user roles deletion](https://app.clickup.com/t/8694mneh7)

src/webapp/components/import-export/ImportTable.tsx
```diff
@@ -0,0 +1,607 @@
...
+            ...defaultUser,
+            id: generateUid(),
+            username: "",
+            password: `District123$`,
+        };
+
+        setUsers(users => users.concat(newUser));
+    }, []);
+
+    let submit: any;
```
eperedo, line 236: I know you're doing this because you need to pass the submit method to the `<ConfirmationDialog />`.

But maybe as an alternative we can use the `Dialog` component and then use a submit button to trigger the action.

```tsx
<Dialog open={true} maxWidth={"lg"} fullWidth={true}>
  <TableContainer>
      {/*  add id to the form */}
      <form id="import-form" onSubmit={handleSubmit}>
         ...
      </form>
  </TableContainer>
  {/* we pass the form id to the form attribute */}
  <Button type="submit" form="import-form">
    Import
  </Button>
</Dialog>
```
I know is more code but it's kinda confusing to have a submit variable and reassigned later inside a component. What do you think?

src/webapp/components/import-export/ImportTable.tsx
```diff
@@ -0,0 +1,607 @@
...
+                    </DialogTooltip>
+                )}
+            </React.Fragment>
+        );
+    };
+
+    const onSubmit = useCallback(
+        async ({ users }: { users: User[] }) => {
+            loading.show(true, i18n.t("Importing users"));
+            const { error } = await onSave(users);
```
eperedo, line 211: We can use the composition root here, right? Instead passing the method as a prop.

eperedo (changes requested):
Thanks @delemaf! Take a look to my comments and let me know if you have any questions. Also I'm having this problem with some fields not showing up. 

image

The csv file I was trying to import is attached in the issue.

src/legacy/models/userHelpers.js
```diff
@@ -49,9 +49,9 @@ const modelByField = {
 };
 
 const queryFieldsByModel = {
-    userRoles: ["id", "displayName"],
-    userGroups: ["id", "displayName"],
-    organisationUnits: ["id", "path", "code", "displayName", "shortName"],
+    userRoles: ["id", "name"],
```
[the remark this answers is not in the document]
delemaf: `UserRoleGroupFF` and `OrgUnitSelectorFF` needed `name` to display

src/webapp/components/import-export/ImportTable.tsx
delemaf: I can call it here but I've seen that ImportTable was also used for in `ReplicateUserFromTable.component`, that's why title and other props are configurable. Should I call composition root here and make it more generic when we migrate `ReplicateUserFromTable` ?

src/webapp/components/import-export/ImportTable.tsx
delemaf: You're right, I think it would be clearer and simpler to use material-ui `Dialog`, `ConfirmationDialog` is not the best when using form

delemaf:
I fixed email, phone number, openId fields. I've refactored some part that you reviewed 👌

eperedo (changes requested):
Thanks @delemaf, not sure if I'm doing something wrong with the csv file, but I'm getting a 409 error because the users array is empty in the request body:

image

can you please take a look and let me know if you can reproduce the error?

Also regarding your questions:

`When importing file, dialog open blank before useGetAllUsers load, I use loading.show() but didn't manage to make it work`

In the `ImportExport.tsx` component we're calling the `onImport` method first, then hiding the loading, that's why you cannot show it again later in the `ImportTable` component. Proposal:

```tsx
const importFromFile = () => { ... })
    .then(result => {
        loading.hide();
        closeMenu();
        onImport(result);
    })
    .catch(err => {
        snackbar.error(err.toString());
        // we can create a function hideLoadingAndClose() with 
        // loading.hide and closeMenu to avoid repeting code
        loading.hide();
        closeMenu();
    });
};
```

`On import user doesn't seems to update but same thing happen in bulk user update so it's maybe my instance`

You can reload the table using the method `this.reloadTable();` in the `List.component.js` file.

```js
_importUsers = async users => {
    const { error } = await this.props.compositionRoot.users.import({ users }).runAsync();
    if (!error) {
        ... 
        this.reloadTable();
        return null;
    } else {
        return error;
    }
};
```

PR #297, opened 2024-06-28
add d2-logger after saving users

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8694wqxd5

### :memo: Implementation

- [x] Settings page for tracker program. 
- [x] Save settings in datastore. Namespace `user-extended-app` key `LOGGER`
- [x] Integrate d2-logger after saving users

**Settings icon at the bottom of the page**
image

**D2-Logger Settings Page**
image

**Events in tracker**
image

### :fire: Testing

- In order to setup d2-logger you will need a tracker program with: 1 attribute of type text to save the username, 1 program stage with 2 dataElements. A dataElement of type FILE and another one of type DATE/DATETIME. (you can see a metadata example in the issue)

- Click in the ⚙️ icon at the bottom of the page
- Select the tracker program you want to use for d2-logger
- Configure the attribute and dataElements.
- Click Save.
- Import some users from a csv file. Then go to the tracker capture app to check the generated logs.

#8694wqxd5


ifoche:
Task linked: [CU-8694wqxd5 [User-Extended] Debug Mode](https://app.clickup.com/t/8694wqxd5)

ifoche (changes requested):
@eperedo I have a couple of comments that I think should be addressed before merging this code:
- The settings access, that you did in a separate button at the botton, could you please just include it in the current existing setting page?
screenshots_1720261211_1220_06072024_667x246
- The settings page seems to be accessible by every user, but the dropdowns are empty and the checkbox is not acting if the user is not a superuser. Couldn't we just show the section in the settings page in grey and disabled if the user is not a superuser? That way there's no confusion. It could also be better just hide it, but I think both would work.
- Could we replace the checkbox "Enable logger" by a slider and if the slider is disabled the following fields are not there?
- I tried to merge with development after having merged a previous PR and it seems like some tests now don't run. Could you merge development code here and fix that?

Thank you!

ifoche:
as an additional comment, I see that the username seems to be filtered by text-type data elements, right? Couldn't it just be a username-type data element?

ifoche:
And.... a last comment. I've checked the json file that's being tracked. I'm not sure that it provides all the information that we actually need. We're trying to track how it is happening this lost of groups. If I'm right, I think that requires tracking what we send and what the system responds to that. I think we're with this PR only tracking what we receive from the system. 

```
// 20240706133800
// https://dev.eyeseetea.com/who-dev-238/api/events/files?eventUid=G3qrMq14epz&dataElementUid=C7KZmliwdOg

{
  "httpStatus": "OK",
  "httpStatusCode": 200,
  "status": "OK",
  "response": {
    "responseType": "ImportReport",
    "status": "OK",
    "stats": {
      "created": 0,
      "updated": 6,
      "deleted": 0,
      "ignored": 0,
      "total": 6
    },
    "typeReports": [
      {
        "klass": "org.hisp.dhis.user.User",
        "stats": {
          "created": 0,
          "updated": 1,
          "deleted": 0,
          "ignored": 0,
          "total": 1
        },
        "objectReports": [
          
        ]
      },
      {
        "klass": "org.hisp.dhis.user.UserGroup",
        "stats": {
          "created": 0,
          "updated": 5,
          "deleted": 0,
          "ignored": 0,
          "total": 5
        },
        "objectReports": [
          
        ]
      }
    ]
  }
}
```

If I see this log, I can only understand that the user updated those groups and a user, but I cannot debug what actually the user did. Could we also send the request? Like having a DE for the request and one for the response?

eperedo:
As we discussed in slack, we're saving metadata and response in different events. Right now there's a bug in Tracker Capture that prevents to download the file for some events if you use the `Timeline`. Changing to `Tabular` fix the problem.

[tracker_tab_entry_error.webm](https://github.com/EyeSeeTea/user-extended-app/assets/461124/34a61429-3682-45e1-9a5e-027dd93f7981)

ifoche (changes requested):
Hi @eperedo . 

Sorry to come back to you on this but it looks like there's one comment that was not implemented yet. I have tried to save the logger settings using a non-admin user and I can change it without any problem. Could you please check it again?
screenshots_1721935043_2117_25072024_1296x548

Thank you!

eperedo:
@ifoche the definition of a super user is: `a user which have the "ALL" authority`. Is this correct? Because I'm reusing the implementation [we already had in the code](https://github.com/EyeSeeTea/user-extended-app/blob/development/src/domain/entities/User.ts#L96). 

I've tested and it works
image

Maybe there's a different concept to consider a user as `superuser`?

ifoche (approved):
thanks @eperedo

PR #298, opened 2024-07-04
saving user groups using the collection endpoint instead metadata

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8694um027

### :memo: Implementation

- [x] Getting users groups from /api/usersGroups instead /api/users
- [x] Saving user groups using the [collection object endpoint](https://docs.dhis2.org/en/full/develop/dhis-core-version-238/developer-manual.html#webapi_adding_removing_objects_collections) instead metadata.

### DIFF
--- a/src/data/repositories/UserD2ApiRepository.ts
+++ b/src/data/repositories/UserD2ApiRepository.ts
@@ -167,13 +167,60 @@ export class UserD2ApiRepository implements UserRepository {
                     })
                 ).flatMap(({ objects }) => {
                     const users = objects.map(user => this.toDomainUser(user));
-                    return this.getLocales(users);
+                    return this.getUsersGroups(usersIds).flatMap(d2UsersWithGroups => {
+                        const usersWithGroups = this.addGroupsToUsers(users, d2UsersWithGroups);
+                        return this.getLocales(usersWithGroups);
+                    });
                 });
             },
             50
         );
 
         return $requests.map(_.flatten);
+    }
+
+    private addGroupsToUsers(users: User[], d2UsersWithGroups: D2UserGroupByKey): User[] {
+        return users.map((user): User => {
+            const userGroups = d2UsersWithGroups[user.id] || [];
+            return { ...user, userGroups: userGroups };
+        });
+    }
+
+    private getUsersGroups(usersIds: Id[]): FutureData<D2UserGroupByKey> {
+        const $requests = chunkRequest(usersIds, usersChunksIds => {
+            return apiToFuture(
+                this.api.models.userGroups.get({
+                    filter: { "users.id": { in: usersChunksIds } },
+                    fields: { id: true, displayName: true, users: true },
+                    paging: false,
+                })
+            ).map(response => {
+                return response.objects.map(d2Group => ({
+                    id: d2Group.id,
+                    name: d2Group.displayName,
+                    users: d2Group.users,
+                }));
+            });
+        });
+
+        return $requests.map(d2UsersGroups => {
+            const userGroups = _(d2UsersGroups)
+                .flatMap(group => group.users.map(user => ({ userId: user.id, group })))
+                .value();
+
+            const groupedByUser = _(userGroups)
+                .groupBy(ug => ug.userId)
+                .mapValues(groups =>
+                    groups.map(g => {
+                        /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
+                        const { users, ...groupWithoutUsers } = g.group;
+                        return groupWithoutUsers;
+                    })
+                )
+                .value();
+
+            return groupedByUser;
+        });
     }
 
     private getFullUsers(options: ListOptions): FutureData<ApiUser[]> {
@@ -198,7 +245,15 @@ export class UserD2ApiRepository implements UserRepository {
                 v: +new Date().getTime(),
             })
         );
-        return userData$.map(({ objects }) => objects);
+        return userData$.flatMap(({ objects }) => {
+            return this.getUsersGroups(objects.map(user => user.id)).map(d2UsersGroups => {
+                const users = objects.map(user => {
+                    const userGroups = d2UsersGroups[user.id] || [];
+                    return { ...user, userGroups };
+                });
+                return users;
+            });
+        });
     }
 
     public listAll(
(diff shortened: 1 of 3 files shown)


ifoche:
And.... a last comment. I've checked the json file that's being tracked. I'm not sure that it provides all the information that we actually need. We're trying to track how it is happening this lost of groups. If I'm right, I think that requires tracking what we send and what the system responds to that. I think we're with this PR only tracking what we receive from the system. 

```
// 20240706133800
// https://dev.eyeseetea.com/who-dev-238/api/events/files?eventUid=G3qrMq14epz&dataElementUid=C7KZmliwdOg

{
  "httpStatus": "OK",
  "httpStatusCode": 200,
  "status": "OK",
  "response": {
    "responseType": "ImportReport",
    "status": "OK",
    "stats": {
      "created": 0,
      "updated": 6,
      "deleted": 0,
      "ignored": 0,
      "total": 6
    },
    "typeReports": [
      {
        "klass": "org.hisp.dhis.user.User",
        "stats": {
          "created": 0,
          "updated": 1,
          "deleted": 0,
          "ignored": 0,
          "total": 1
        },
        "objectReports": [
          
        ]
      },
      {
        "klass": "org.hisp.dhis.user.UserGroup",
        "stats": {
          "created": 0,
          "updated": 5,
          "deleted": 0,
          "ignored": 0,
          "total": 5
        },
        "objectReports": [
          
        ]
      }
    ]
  }
}
```

If I see this log, I can only understand that the user updated those groups and a user, but I cannot debug what actually the user did. Could we also send the request? Like having a DE for the request and one for the response?

ifoche (approved):
approved from a functional perspective, but waiting for a technical review from @tokland

src/data/repositories/UserD2ApiRepository.ts
```diff
@@ -176,6 +179,50 @@ export class UserD2ApiRepository implements UserRepository {
...
+        return $requests.map(d2UsersGroups => {
+            const userGroups = _(d2UsersGroups)
+                .flatMap(group => group.users.map(user => ({ userId: user.id, group })))
+                .value();
+
+            const groupedByUser = _(userGroups)
+                .groupBy(ug => ug.userId)
+                .mapValues(groups =>
+                    groups.map(g => {
+                        /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
```
tokland, line 215: To avoid "eslint-disable", 2 ways:

- `const { users: _users, ...groupWithoutUsers } = g.group;`
- `const groupWithoutUsers = _.omit(g.group, ["users"]); `

src/data/repositories/UserD2ApiRepository.ts
```diff
@@ -330,42 +386,60 @@ export class UserD2ApiRepository implements UserRepository {
...
+
+    private buildGroupsToSave(userGroup: { id: Id; users: Array<{ id: Id }> }, action: D2ActionGroup) {
+        const isAdding = action === "add";
+        const usersIds = userGroup.users.map(({ id }) => ({ id: id }));
+        return apiToFuture(
+            this.api.request({
+                method: "post",
+                url: `/userGroups/${userGroup.id}/users`,
+                data: isAdding ? { additions: usersIds } : { deletions: usersIds },
+            })
```
tokland, line 442: Do we need to check the response of the POST to see if it was successful?

src/legacy/models/userHelpers.js
```diff
@@ -381,14 +382,17 @@ async function getUserGroupsToSaveAndPostMetadata(d2, api, users, existingUsersT
...
+    const userRepository = new UserD2ApiRepository({ url: d2Api.baseUrl });
     const existingUsersToUpdate = await getExistingUsers(d2, {
         fields: ":owner,userCredentials,userGroups[id]",
         filter: "userCredentials.username:in:[" + _(users).map("username").join(",") + "]",
     });
     const usersToSave = getUsersToSave(users, existingUsersToUpdate);
-    return getUserGroupsToSaveAndPostMetadata(d2, api, usersToSave, existingUsersToUpdate);
+    const response = await postMetadata(api, { users: users });
+    await userRepository.getGroupsToSave(usersToSave, existingUsersToUpdate).runAsync();
+    return response;
```
tokland, line 395: Clean arch was already poorly implemented for the whole file, and probably it's not the moment to fix it, but some notes:

- Name `getGroupsToSave` seems misleading, as it's not a get, it's an update action. Maybe `updateUserGroups`?
- I see that getUserGroupsToSaveAndPostMetadata is still used "saveCopyInUsers", is that a problem?

tokland (commented):
[code-only review]

src/legacy/models/userHelpers.js
eperedo: Not a problem because `Copy in user` was already migrated to typescript + CA by Florian.

src/data/repositories/UserD2ApiRepository.ts
```diff
@@ -231,19 +280,20 @@ export class UserD2ApiRepository implements UserRepository {
...
+                    const user = users.find(user => user.id === existingUser.id);
+                    if (!user) return undefined;
+                    return this.buildUsersToSave(existingUser, user);
+                })
+                .compact()
+                .value();
+
+            return apiToFuture(this.api.metadata.post({ users: usersToSend })).flatMap(data => {
+                return Future.joinObj({
+                    saveLocales: this.saveLocales(usersToSave).map(() => data),
```
tokland, line 294: I was that `.map(() => data)` was a left-over when refactoring, it's not needed here.

src/data/repositories/UserD2ApiRepository.ts
```diff
@@ -330,43 +380,83 @@ export class UserD2ApiRepository implements UserRepository {
...
+        ).flatMap(d2Response => {
+            const response = d2Response.response ? d2Response.response : d2Response;
+            const errorMessage = getErrorFromResponse(response.typeReports);
+            if (response.status === "ERROR") return Future.error(errorMessage);
+            return Future.success(
+                new Stats({
+                    created: response.stats.created,
+                    updated: response.stats.updated,
+                    ignored: response.stats.ignored,
+                    deleted: response.stats.deleted,
```
tokland, line 455: Unless you prefer to be explicit: `...response.stats`

tokland (approved):
code-wise, it looks good to me. I added a pair of comments, but very minor, approving.

PR #304, opened 2024-11-19
Dojo: Replicate Action Refactor

nshandra:
### :pushpin: References

-   **Issue:** [Q4 2024 Project Coding Dojo](https://app.clickup.com/t/8695pt0av)

### :memo: Implementation

Refactor replicate functions:

- Replicate user from template (A final form version is available and will probably be the one to integrate into UEApp)
- Replicate user from table (Updated the new ImportTable for this function)


nshandra:
@xurxodev putting on draft to remove the demo part from the PR

src/data/repositories/UserD2ApiRepository.ts
```diff
@@ -152,6 +152,22 @@ export class UserD2ApiRepository implements UserRepository {
         ).map(({ objects }) => objects.map(user => user.id));
     }
 
+    public listAllUsernames(options: ListOptions): FutureData<string[]> {
```
xurxodev: A repository should never return primitive types, this is a code smell that we saw in the coding dojo

In Clean Architecture, a repository should always return entities.

The presentation layer should map user list to get userNames if required.

In this case the entity is UserLogic (can we change to User and User to UserProps?)

src/domain/usecases/ListAllUsernamesUseCase.ts
```diff
@@ -0,0 +1,11 @@
+import { UseCase } from "../../CompositionRoot";
```
xurxodev: A use case should never return primitive types, this is a code smell that we saw in the coding dojo

In Clean Architecture, a use case should always return entities.

The presentation layer should map user list to get userNames if required.

In this case the entity is UserLogic (can we change to User and User to UserProps?)

src/webapp/components/import-export/ImportTable.tsx
```diff
@@ -597,14 +630,23 @@ const useValidations = (
             return {
                 validation: (value: string) => {
                     if (!value) return i18n.t("Please provide a username");
+                    if (/^[._@-]|[._@-]$/.test(value)) {
```
xurxodev: These new validations are business rules validations and should be realized in the user domain entity

src/webapp/components/replicate/ReplicateUserFromTable.tsx
```diff
@@ -0,0 +1,122 @@
...
+import { useLoading, useSnackbar } from "@eyeseetea/d2-ui-components";
+import { generateUid } from "../../../utils/uid";
+
+interface ReplicateUserFromTableProps {
+    userToReplicateId: Id;
+    onRequestClose: () => void;
+}
+
+export const ReplicateUserFromTable: React.FC<ReplicateUserFromTableProps> = props => {
+    const { compositionRoot } = useAppContext();
```
xurxodev: The component should only render. (render logic)

The rest of the presentation logic (manage state, actions, invoke use cases, etc.) should be in a custom hook: useReplicateUserFromTable.

In this custom hook, we should use the user entity, not the user props

src/webapp/components/replicate/ReplicateUserFromTemplate.tsx
```diff
@@ -0,0 +1,368 @@
...
+import { Id } from "../../../domain/entities/Ref";
+import { User, defaultUser } from "../../../domain/entities/User";
+import { UserLogic } from "../../../domain/entities/UserLogic";
+
+interface ReplicateUserFromTemplateProps {
+    userToReplicateId: Id;
+    onRequestClose: () => void;
+}
+
+export const ReplicateUserFromTemplate: React.FC<ReplicateUserFromTemplateProps> = props => {
```
xurxodev, line 16: The rest of the presentation logic (manage state, actions, invoke use cases, etc.) should be in a custom hook: useReplicateUserFromTemplate.

In this custom hook, we should use the user entity, not the user props

src/webapp/components/replicate/ReplicateUserFromTemplate.tsx
```diff
@@ -0,0 +1,368 @@
...
+                        <Button disabled={hasValidationErrors} type="submit" form="replicate-form" color="primary">
+                            {i18n.t("Replicate")}
+                        </Button>
+                    </DialogActions>
+                </Dialog>
+            )}
+        </>
+    );
+};
+
```
xurxodev, line 78: These validations are business rules validations and should be realized in the user domain entity.

To validate you can to try to create a domain entity, call internally to validate a return, for example an Either with the entity or the errors.

You can see these examples:
* https://github.com/EyeSeeTea/malaria-threats-map/blob/c69d08e377d83d56f2d08da54899616b97f5492c/src/domain/entities/Feedback.ts#L36
* https://github.com/EyeSeeTea/coding-dojo/blob/kata/clean-architecture-solution/src/domain/entities/Product.ts

xurxodev (changes requested):
thanks @nshandra 

I've found issues to review:

- Non related specially to this PR but UserLogic is really the entity and User is the props. I would expect the names as UserLogic -> User and User -> UserProps
- In the new code affected by this PR that to use  UserRepository, we should change to return the entity from the repository and not the props.

I've added rest of the issues in comments in the code.

src/data/repositories/UserD2ApiRepository.ts
nshandra: With the proposed changes to `listAllUsernames` and `ListAllUsernamesUseCase.ts` they do the same than `listAllUsers` / `ListAllUsersUseCase`. I switched to using them.

src/domain/usecases/ListAllUsernamesUseCase.ts
nshandra: ListAllUsernamesUseCase was removed.

src/webapp/components/import-export/ImportTable.tsx
nshandra: Moved to User (former UserLogic) entity

src/webapp/components/replicate/ReplicateUserFromTable.tsx
nshandra: Moved logic to `useReplicateUserFromTable.ts`

src/webapp/components/replicate/ReplicateUserFromTemplate.tsx
nshandra: Moved logic to `useReplicateUserFromTemplate.ts`

src/webapp/components/replicate/ReplicateUserFromTemplate.tsx
nshandra: Created a new entity `ReplicateTemplate` with the validation logic and some improvements to them.

src/data/D2ApiLogger.ts
```diff
@@ -2,7 +2,7 @@ import _ from "lodash";
 import { TrackerProgramLogger } from "@eyeseetea/d2-logger";
 import { Future, FutureData } from "../domain/entities/Future";
 import { LoggerSettings } from "../domain/entities/LoggerSettings";
-import { User } from "../domain/entities/User";
+import { UserProps } from "../domain/entities/UserProps";
```
xurxodev: UserProps should be used outside of the domain. 

You can use User because extends Struct<UserProps> and has the same properties

src/data/repositories/UserD2ApiRepository.ts
```diff
@@ -30,30 +30,30 @@ export class UserD2ApiRepository implements UserRepository {
         this.userStorage = new DataStoreStorageClient("user", instance);
     }
```
xurxodev: A repository always must receive and return entities, not props

src/domain/entities/ReplicateTemplate.ts
```diff
@@ -0,0 +1,112 @@
...
+export class ReplicateTemplate extends Struct<ReplicateTemplateProps>() {
+    private static indexString = "$index";
+    private static minCount = 1;
+    private static maxCount = 100;
+
+    static getFromTemplate(template: string, index: number): string {
+        return template.replace(/\$index/g, index.toString());
+    }
+
+    static validateReplicateCount(value: string): string | undefined {
```
xurxodev, line 41: An entity should validate its invariants in the moment of the construction.

When an entity doesn't validate its invariants, is possible to create an instance in invalid state

the validations should be executed to create or update the entity

src/domain/entities/User.ts
```diff
@@ -1,105 +1,169 @@
...
+        return language || "en";
+    }
+
+    static validateHasRequiredFields(users: UserProps[]): boolean {
+        return users.every(
+            user => user.organisationUnits.length > 0 && user.userRoles.length > 0 && user.userGroups.length > 0
+        );
+    }
+
+    static validateUniqueOpenId(users: UserProps[]): boolean {
```
xurxodev, line 99: An entity should validate its invariants in the moment of the construction.

When an entity doesn't validate its invariants, is possible to create an instance in invalid state

the validations should be executed to create or update the entity

src/domain/repositories/UserRepository.ts
```diff
@@ -3,20 +3,20 @@ import { FutureData } from "../entities/Future";
...
 import { NamedRef } from "../entities/Ref";
 import { Stats } from "../entities/Stats";
-import { User } from "../entities/User";
+import { UserProps } from "../entities/UserProps";
 
 export interface UserRepository {
-    getCurrent(): FutureData<User>;
-    list(options: ListOptions): FutureData<PaginatedResponse<User>>;
-    listAll(options: ListOptions): FutureData<User[]>;
+    getCurrent(): FutureData<UserProps>;
```
xurxodev: A repository always must receive and return entities, not props

src/domain/usecases/CopyInUserUseCase.ts
```diff
@@ -21,30 +21,30 @@ export class CopyInUserUseCase {
         });
     }
 
-    private applyCopyToUsers(usersToUpdate: User[], options: CopyInUserOptions): User[] {
```
xurxodev: An use case can receive the entity or the props, but if receive the props normally is in creation or modifications, then the use case should create the entity but never pass it directly to repositories.

src/domain/usecases/ExportUsersUseCase.ts
```diff
@@ -57,7 +57,7 @@ export class ExportUsersUseCase {
     }
 
     private buildBlobAndFilename(
-        users: User[],
```
xurxodev: An use case can receive the entity or the props, but if receive the props normally is in creation or modifications, then the use case should create the entity but never pass it directly to repositories.

src/domain/usecases/GetCurrentUserUseCase.ts
```diff
@@ -1,12 +1,12 @@
 import { UseCase } from "../../CompositionRoot";
 import { FutureData } from "../entities/Future";
-import { User } from "../entities/User";
+import { UserProps } from "../entities/UserProps";
 import { UserRepository } from "../repositories/UserRepository";
 
 export class GetCurrentUserUseCase implements UseCase {
```
xurxodev: The use case should return always the entity

src/domain/usecases/GetUsersByIdsUseCase.ts
```diff
@@ -1,12 +1,12 @@
 import { UseCase } from "../../CompositionRoot";
 import { FutureData } from "../entities/Future";
-import { User } from "../entities/User";
+import { UserProps } from "../entities/UserProps";
 import { UserRepository } from "../repositories/UserRepository";
 
 export class GetUsersByIdsUseCase implements UseCase {
     constructor(private userRepository: UserRepository) {}
 
-    public execute(ids: string[]): FutureData<User[]> {
```
xurxodev: The use case should return always the entity

src/domain/usecases/ListUsersUseCase.ts
```diff
@@ -1,13 +1,13 @@
...
 import { UseCase } from "../../CompositionRoot";
 import { FutureData } from "../entities/Future";
-import { User } from "../entities/User";
+import { UserProps } from "../entities/UserProps";
 import { UserRepository, ListOptions } from "../repositories/UserRepository";
 
 export class ListUsersUseCase implements UseCase {
     constructor(private userRepository: UserRepository) {}
 
-    public execute(options: ListOptions): FutureData<{ pager: Pager; objects: User[] }> {
```
xurxodev: The use case should return always the entity

src/domain/usecases/RemoveUsersUseCase.ts
```diff
@@ -1,12 +1,12 @@
...
 import { Stats } from "../entities/Stats";
-import { User } from "../entities/User";
+import { UserProps } from "../entities/UserProps";
 import { UserRepository } from "../repositories/UserRepository";
 
 export class RemoveUsersUseCase {
     constructor(private userRepository: UserRepository) {}
 
-    execute(users: User[]): FutureData<Stats> {
+    execute(users: UserProps[]): FutureData<Stats> {
```
xurxodev: To do this use case more usable from several places is better if receive only userIds, to remove is not necessary all props or entity

src/domain/usecases/ReplicateFromTemplateUseCase.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export class ReplicateFromTemplateUseCase implements UseCase {
+    constructor(private userRepository: UserRepository) {}
+    execute(
+        sourceUser: UserProps,
+        count: number,
+        usernameTemplate: string,
+        passwordTemplate: string
+    ): FutureData<MetadataResponse> {
+        const newUsers: UserProps[] = _.times(count, index => {
```
xurxodev: To replicate a user, we should try to create the entity and to realize the validations in the moment of the creation (using create or build method in the entity with validations)

src/webapp/components/import-export/ImportTable.tsx
```diff
@@ -596,66 +611,76 @@ const useValidations = (
...
                         return i18n.t("User already exists");
-                    } else {
-                        const validators = composeValidators(
-                            string,
-                            createMinCharacterLength(2),
-                            createMaxCharacterLength(140)
-                        );
-                        return validators(value);
                     }
+                    const usernameValidationError = User.validateUsername(value);
```
xurxodev, line 618: The validations should be realized in the moment of the creation. It's ok if you want to use the function to validate here instead of try to create the entity but it's necessary to validate again to create the entity to avoid invalid states

xurxodev (changes requested):
thanks @nshandra 

This PR has become quite large.

I've noticed some important issues in the implementation of Clean Architecture and decided to focus on that first. I've left notes about them in the comments.

xurxodev (approved):
thanks @nshandra 

In general terms, the code of this PR has improved. There are architecture improvements to do, but as this PR is blocking others, I leave it here for the future.

@adrianq This improvements can be different tasks

- Validations return primitive types like strings. booleans.

Validations should represents an error to avoid interpretations. 
To use throw error or either

- Duplicated validations

There are duplicate validations like username, etc.

Username, email and similar concepts are really domain types and it should modeled using ValueObjects and not primitive types to avoid the primitive obsession code smells and to reuse this types and its validations in different parts of the app

- A lot of static validations 

There are a lot of static public validations that never are called from outside of the entity like ValidateEmail for example.

This maybe solved using value objects

- generateUid() duplicate in use cases and custom hook

Generate uid should be realized only from the domain and to avoid duplicates the best place in in the user domain entity to create.

If the uid is empty then generate it.

* Performance

I've done a simple test of replicate from table and replicate from template, and I have the feeling the poor performance to load the dialogs and to save.
The save in my case is to save one user to replicate.

adrianq:
Thanks @xurxodev  I have created two tasks to cover your comments:
- This task is regarding the refactors on the code: https://app.clickup.com/t/869af8g88
- This other task is about the performance and is the one with the highest performance: https://app.clickup.com/t/869af8h25 @nshandra has already talked to the two of us and has some ideas about how to improve it. We are merging this PR. Some other PRs will be merged afterwards that include the schemas for 41 so it is much more convenient to start from a new branch from development

PR #307, opened 2025-01-14
Feature: Add BundleMon action to check the build size changes

nshandra:
### :pushpin: References

-   **Issue:** [GitHub Action size](https://app.clickup.com/t/8697f4ne3)

### :memo: Implementation

GitHub action to monitor the build and/or resulting .zip file size.

General steps:
1. [Install BundleMon GitHub App](https://github.com/apps/bundlemon)
2. Create .bundlemonrc via: `size_limit="X MB" percent_limit="Y" yarn make-bundlemonrc` or copy the `.bundlemonrc_template` and edit manually. This file will define the report.
3. Create a workflow file for bundlemonrc.

####  yarn build
This PR edits the build action to copy the ZIP file to the build folder. This is needed because BundleMon has trouble processing large repositories (the ZIP is in the root of the repo, this means that baseDir is '.', leading to this error).
If the ZIP file size is not monitored, this change can be discarded.

####  yarn make-bundlemonrc
This PR also adds a new yarn action yarn make-bundlemonrc that will create a .bundlemonrc file from the .bundlemonrc_template.
This command needs to be provided with vars to convert the template into an usable file:
  - The size limit is a string that consists of a number and unit of measurement (KB, MB...)
  - The percent limit is a number of the maximum size increase allowed (in %)

If the ZIP file size is not monitored, the $build_file_types generation can be removed.

#### Template and Demo

Originally the template included redundant entries as a demo, an example of how the look like can be seen [here](https://github.com/EyeSeeTea/security-tests/pull/3#issuecomment-2589431579).

The template used there looks like this (this one is adapted to UEApp):
```
{
    "baseDir": "./build",
    "files": [
        // Monit the build ZIP file.
        {
            "path": "./$npm_package_name.zip",
            "friendlyName": "Build ZIP",
            "compression": "none",
            "maxSize": "$size_limit",
            "maxPercentIncrease": "$percent_limit"
        },
        // Monit a specific file with a "dynamic" name (eg: main.87c4a820.chunk.js)
        // If the <hash> is not specified correctly BundleMon will report the old name as a deletion and the new one as a new file.
        {
            "path": "./static/js/main.<hash>.chunk.js",
            "friendlyName": "Main",
            "compression": "none",
            "maxPercentIncrease": "$percent_limit"
        },
        // Monit all files from baseDir except ZIP type.
        // The idea is to have a list of the files so its easy to identify the offending file.
        // In practice this results in a overly verbose message or in false positives deletions due to dynamic names.
        // Its better to use the CLI to find the file.
        {
            "path": "./**/*.{$build_file_types}",
            "friendlyName": "Build Folder Files",
            "compression": "gzip"
        }
    ],
    // Monit the build folder as a whole. This is an alternative to copying the ZIP file to the build folder.
    // Note that gzip will produce slightly smaller files than zip.
    "groups": [
        {
           // Unfortunately, looks like there is no way of using a negation operator here
            "path": "./**/*.{$build_file_types}",
            "friendlyName": "Build Folder",
            "compression": "gzip",
            "maxSize": "3 MB",
            "maxPercentIncrease": 25
        }
    ],
    "reportOutput": [
        [
            "github",
            {
                "checkRun": true,
                "commitStatus": true,
                "prComment": true
            }
        ]
    ]
}
```

BundleMon can be executed locally in the repository via:
`yarn bundlemon`

If a specific config file needs to be used use the `-c`,`--config` option:
`yarn bundlemon --config conf_file`

This is useful to validate the config without waiting for the action to run. 

### Docs:
https://github.com/LironEr/bundlemon
https://github.com/LironEr/bundlemon-github-actions
https://github.com/LironEr/bundlemon-action


package.json
```diff
@@ -118,7 +119,8 @@
         "cy:e2e:open": "NODE_ENV=test CYPRESS_E2E=true cypress open",
         "cy:e2e:run": "NODE_ENV=test CYPRESS_E2E=true cypress run --browser chrome",
         "prepare": "husky install",
-        "script-example": "npx ts-node src/scripts/example.ts"
+        "script-example": "npx ts-node src/scripts/example.ts",
+        "make-bundlemonrc": "build_file_types=$(find ./build -type f ! -name '*.zip' | sed -n 's/.*\\.//p' | sort -u | paste -sd,) && build_file_types=$build_file_types envsubst < .bundlemonrc_template > .bundlemonrc"
```
tokland: Typically, we prefer scripts that run cross-platform (Linux/Mac/Windows), and then we would write a TS script. But this being a script that's only used once, I don't think it's a big deal, we can keep Bash.

But that I would do: it's a bit long and difficult to read. Maybe create a sh script and call it? and we can use best practices in that script (and now we can have default values for the size variables, now if the user forgets, they are empty and wrong on the config file)

.bundlemonrc_template
```diff
@@ -0,0 +1,31 @@
+{
+    "baseDir": "./build",
+    "files": [
+        {
+            "path": "./$npm_package_name.zip",
```
tokland: > If the ZIP file size is not monitored, this change can be discarded

Unless it's something you already agreed with @adrianq, I don't think we gain much from checking both the ZIP and the contents zipped. I'd check just the contents. As a bonus, we could remove all the gymnastics required by the zip being on folder build/.

tokland (commented):
Just some comments about minor implementation details. 

I tested it in skeleton-app (my fork) and it worked nicely, so it also works with vite out of the box: https://github.com/tokland/dhis2-app-skeleton/pull/2

.bundlemonrc_template
nshandra: I included both at the moment of creating the PR to show the options.
Adrian mentioned that the group method was fine. I will modify the PR accordingly.

package.json
nshandra: Given that the config is simple with the Build Folder, all the changes to package were removed.

tokland (approved):
Looks good and simple :+1:

adrianq:
Superseeded by https://github.com/EyeSeeTea/user-extended-app/pull/314 Thanks @nshandra  This was the first step in the right direction

PR #285, opened 2024-05-15
advanced filter for active and inactive users

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8694jh1ge

### :video_camera: Screenshots/Screen capture

https://github.com/EyeSeeTea/user-extended-app/assets/37223065/0be0de22-16f9-4b4a-8e65-c6f9fddca1da

### :fire: Testing

#8694jh1ge

PR #308, opened 2025-01-29
Delimit users list and add reset password

p3rcypj:
### :pushpin: References

- **Issue:** https://app.clickup.com/t/869656vnq

### :memo: Implementation

PRs directly affected:
- #285
  - ➡️ Overriding filter, dropdown disabled state.
  - ❌ (Not done) I would change dropdown to another `SegmentedControl` with `ACTIVE | INACTIVE | BOTH`
- #233 and #234 
  - ➡️ Override as "AND", changed to `SegmentedControl` (UX decision), moved warning to info icon
  - ❌ (Not done) I would remove warning because I don't see the confusion between query + filter with `rootJunction`, but maybe I'm not considering enough user's context knowledge. Pinging @ifoche and @tokland to request approval for moving the warning text to the info icon (see attached screenshots below) or even remove it.
- #287, #289, #295
  - ➡️ Requests to all users clamped with the new filters

Subtasks:
- Add setting to hide feedback
- Add setting to show only users assigned to users' organisation units (include children as DHIS2 does in its Users page/app)
- Add setting to show only active users
- Add reset password to actions (only for users with email)
- Override active filter and hide "Display users I can manage" if "Show only active users" option is truthy
- If "Show only active users" is set, root junction will be forced to be "AND" and dropdown will be disabled in "Active" state to prevent user to change it.

Minor features:
- AppSettings into AppSettingsContext (allowing changes without a refresh)
- Allowed backdrop to close dialog on settings dialog
- Changed root junction `Switch` to `SegmentedControl` from `@dhis2/ui`
- Added disabled state for dropdown in filters
- Small change to tests because they were failing when comparing the filenames, a seconds increased just before ending test
- Warning for user about root junction was "Only for advanced filters" moved to the info icon in a `title` attribute.

DHIS2 Bug (not on Jira yet):
 - If both "Show only active users" and "Show only users assigned to users' organisation units" filters are set, the endpoint returns a wrong total count, making the pagination to not work correctly. After all "correct" users are returned, next pages just return the "admin" user. Therefore, we manually check the page where the users count is less than the pageSize, making that page the last one and clamping pagination to prevent more "next" requests. Then, the real total is counted and shown. For the previous pages, total count is hidden by targeting with CSS the element and adding a `:before` pseudo-element with the new content text; being a much less invasive patch than editing the component in d2-ui-components

Small note about UI:
- On our legacy components we are using the material-ui 0.20.2 version. On the process on turning our legacy components into TS, I would only first change the components to use `@dhis2/ui`; making the transition to TS a bit less painful.
  - If approved, I would also change the components we are adding with this PR also to `@dhis2/ui`

#### (No screen recording to not show sensitive data from the docker I was using)

image

_Showing overrided "Filter by active users" with "OR" as default_

image

_Showing overrided "Filter by active users" with "AND" as both filters are set, along with hiding the "Display only users I can manage"_

image

_Showing permissions tab on Dialog_

image

_Showing DHIS2 total count bug patch_

### :fire: Testing / Final notes to reviewer

- Please delete your `settings` namespace on dataStorage before testing or add the keys manually. Because moved from `Partial<AppSettings>` to mandatory for better definition and linting.
- After checking the `rootJunction` is only mandatory to be "AND" for the option of active users, making `userOrgUnits` param independent


src/data/repositories/UserD2ApiRepository.ts
```diff
@@ -131,14 +139,33 @@ export class UserD2ApiRepository implements UserRepository {
                 canManage: canManage === "true" ? "true" : undefined,
                 filter: otherFilters,
                 rootJunction: areFiltersEnabled ? rootJunction : undefined,
+                userOrgUnits: onlyUsersOrgUnits === true ? "true" : undefined,
```
eperedo: `onlyUsersOrgUnits` is a boolean we can avoid the explicit evaluation to true.

```ts
onlyUsersOrgUnits === true ? "true" : undefined,
```

src/data/repositories/UserD2ApiRepository.ts
```diff
@@ -147,6 +174,8 @@ export class UserD2ApiRepository implements UserRepository {
                 query: search !== "" ? search : undefined,
                 canManage: canManage === "true" ? "true" : undefined,
                 filter: otherFilters,
+                userOrgUnits: onlyUsersOrgUnits === true ? "true" : undefined,
+                includeChildren: onlyUsersOrgUnits === true ? "true" : undefined,
```
eperedo: I know this is a small check, but since we're already repeating it multiple times extracting into a function and reusing would be a good idea.

src/webapp/components/user-list-table/UserListTable.tsx
```diff
@@ -55,10 +57,13 @@ function convertActionToOrgUnitType(action: ActionType): SaveUserOrgUnitOptions[
...
-        case "copy_in_user":
-        case "disable":
-        case "enable":
-        case "remove":
+        // Note for reviewer: Commented these lines just to keep track of them and because linter was complaining about return statement
+        // If you give me the ok, I would restrict the ActionType to OrgUnitActionType
+        // case "copy_in_user":
+        // case "disable":
+        // case "enable":
+        // case "remove":
```
eperedo: I didn't have any warning with eslint, but yeah restricting to OrgUnitActionType sounds a good idea.

src/webapp/components/user-list-table/UserListTable.tsx
```diff
@@ -311,6 +326,17 @@ export const UserListTable: React.FC<UserListTableProps> = ({
...
+                {
+                    name: "reset_password",
+                    text: i18n.t("Reset password"),
+                    icon: <Icon>lock</Icon>,
+                    multiple: true,
+                    onClick: users => {
+                        setSelectedUserIds(users);
+                        setActionType("reset_password");
+                    },
+                    isActive: checkHasEmail(),
```
eperedo: This is ok, but I think the implementation belongs to the domain (User.ts) and should be reuse it here and it the usecase. That way if anyone wants to reuse the usecase (other part in the UI, a script, etc) we can be sure that all the validations are gonna be executed.

src/webapp/components/users-remove-modal/UsersSelectedModal.tsx
```diff
@@ -21,17 +21,27 @@ export type ActionType =
...
-        return { title: "Remove" };
-    } else if (actionType === "disable") {
-        return { title: "Disable" };
-    } else if (actionType === "enable") {
-        return { title: "Enable" };
+// TODO: Add into i18n translation files (as will not be recognized by i18n-scanner)
+function getMessagesByActionType(actionType: ActionType): { title: string; description: string; success: string } {
+    switch (actionType) {
+        case "remove":
+            return { title: "Remove users", description: "remove", success: "Users removed" };
```
eperedo: Add i18n

eperedo (changes requested):
Thanks @p3rcypj, we're very close. Feel free to reach out in case any questions

src/webapp/components/user-list-table/UserListTable.tsx
p3rcypj: I wanted to do it actually as you mention. But I couldn't because it was the way the component was asking:
```typescript
function checkHasEmail() {
    const currentUserHasUpdateAccessOn = checkAccess(["update"]);

    return (users: User[]) => currentUserHasUpdateAccessOn(users) && _(users).some(user => Boolean(user.email));
}
```
Even `users: User[]` and `users.some` seems weird because what actually ends up is just one user inside an array per row. Like `[{...}]` for each row. On the user it would be something like `hasEmail()` for example, but it would only replace the `Boolean(user.email)` part because the component asks for `(users: User[]) => boolean`.

eperedo: I was thinking on moving this functions to the domain:

```ts
// User.ts
export function checkAccess(requiredKeys: string[]) {
    return (users: User[]) =>
        _(users).every(user => {
            const permissions = _(user.access).pickBy().keys().value();
            return _(requiredKeys).difference(permissions).isEmpty();
        });
}

export function checkHasEmail(users: User[]) {
    const currentUserHasUpdateAccessOn = checkAccess(["update"]);

    return currentUserHasUpdateAccessOn(users) && _(users).some(user => Boolean(user.email));
}
```

Then reusing them in the component and usecase:

```ts
// ResetUsersPasswordsUseCase.ts
execute(users: User[]): FutureData<Stats> {
        const allUsersCanBeUpdated = checkHasEmail(users);
        if (allUsersCanBeUpdated) {
            return this.userRepository.resetPasswords(users);
        } else {
            return Future.error("Not all users have emails");
        }
    }
```

//UserListTable.tsx

````ts
{
                    name: "reset_password",
                    text: i18n.t("Reset password"),
                    icon: <Icon>lock</Icon>,
                    multiple: true,
                    onClick: users => {
                        setSelectedUserIds(users);
                        setActionType("reset_password");
                    },
                    isActive: users => {
                        return checkHasEmail(users);
                    },
                }
```

eperedo (changes requested):
Looks good, but I'm getting an error in  the UsersSelectedModal component related to i18n.

Select any user in the table then click on the `Reset password` action:

```
properties of undefined (reading 'translator')
    at t (i18next.js:318:1)
    at getMessagesByActionType (UsersSelectedModal.tsx:41:1)
    at UsersSelectedModal (UsersSelectedModal.tsx:74:1)
    at renderWithHooks (react-dom.development.js:14985:1)
    at mountIndeterminateComponent (react-dom.development.js:17811:1)
    at beginWork (react-dom.development.js:19049:1)
    at HTMLUnknownElement.callCallback (react-dom.development.js:3945:1)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:3994:1)
    at invokeGuardedCallback (react-dom.development.js:4056:1)
    at beginWork$1 (react-dom.development.js:23964:1)
    at performUnitOfWork (react-dom.development.js:22776:1)
    at workLoopSync (react-dom.development.js:22707:1)
    at renderRootSync (react-dom.development.js:22670:1)
    at performSyncWorkOnRoot (react-dom.development.js:22293:1)
    at react-dom.development.js:11327:1
    at unstable_runWithPriority (scheduler.development.js:468:1)
    at runWithPriority$1 (react-dom.development.js:11276:1)
    at flushSyncCallbackQueueImpl (react-dom.development.js:11322:1)
    at flushSyncCallbackQueue (react-dom.development.js:11309:1)
    at scheduleUpdateOnFiber (react-dom.development.js:21893:1)
    at dispatchAction (react-dom.development.js:16139:1)
    at userHooks.ts:43:1
```

src/webapp/components/user-list-table/UserListTable.tsx
p3rcypj: Ohh I see it now! Makes sense! Good point 💯

eperedo (approved):

PR #309, opened 2025-02-16
Add selector for access to settings

p3rcypj:
### :pushpin: References

-   **Issue:** Closes subtask https://app.clickup.com/t/8697qhzkf (main task https://app.clickup.com/t/869656vnq)

### :memo: Implementation

Added Sharing Settings component from d2-ui-components to define the list of users/user groups with access to the settings interface. If not set, access to settings will be public for everyone. Changes will occur after reloading to prevent users from losing access to settings in case they mistakenly set incorrect permissions.

image
image

### DIFF
--- a/src/domain/entities/Permission.ts
+++ b/src/domain/entities/Permission.ts
@@ -1,0 +1,7 @@
+import { NamedRef } from "./Ref";
+
+export type Permission = {
+    publicAccess: string; // '--------' | 'r-------'  | 'rw------'
+    users: NamedRef[];
+    userGroups: NamedRef[];
+};
--- a/src/domain/entities/Permission.ts
+++ b/src/domain/entities/Permission.ts
@@ -1,7 +1,39 @@
+import _ from "lodash";
+import { Struct } from "./generic/Struct";
 import { NamedRef } from "./Ref";
 
 export type Permission = {
-    publicAccess: string; // '--------' | 'r-------'  | 'rw------'
     users: NamedRef[];
     userGroups: NamedRef[];
 };
+
+type PublicPermissionAttrs = Permission & {
+    publicAccess: AccessValue;
+};
+
+export class PublicPermission extends Struct<PublicPermissionAttrs>() {
+    updateUsers(users: NamedRef[]): PublicPermission {
+        return this._update({
+            users,
+            publicAccess: _.isEmpty(users) && _.isEmpty(this.userGroups) ? AccessValue.public() : AccessValue.private(),
+        });
+    }
+
+    updateUserGroups(userGroups: NamedRef[]): PublicPermission {
+        return this._update({
+            userGroups,
+            publicAccess: _.isEmpty(this.users) && _.isEmpty(userGroups) ? AccessValue.public() : AccessValue.private(),
+        });
+    }
+}
+
+// Posibility of adding write access if needed in future
+class AccessValue extends Struct<{ read: boolean }>() {
+    static public() {
+        return new AccessValue({ read: true });
+    }
+
+    static private() {
+        return new AccessValue({ read: false });
+    }
+}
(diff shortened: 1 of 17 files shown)


src/domain/entities/UserSearch.ts
```diff
@@ -0,0 +1,9 @@
+type UserSearchItem = {
+    id: string;
+    displayName: string;
```
eperedo: `displayName` is a DHIS thing, we should use `name` and then map the value in the implementation. `UserSearchD2Repository`

src/domain/usecases/CheckCurrentUserCanAccessSettingsUseCase.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    execute(): FutureData<boolean> {
+        return Future.joinObj({
+            user: this.userRepository.getCurrent(),
+            appSettings: this.appSettingsRepository.get(),
+        }).map(({ user, appSettings }) => this.checkUserCanAccessSettings(user, appSettings));
+    }
+
+    private checkUserCanAccessSettings(user: User, appSettings: AppSettings): boolean {
+        const permissions = appSettings.settingsAccess;
```
eperedo, line 20: We need a guard here checking if the current user is admin then avoid extra validations, right?

```ts
    private checkUserCanAccessSettings(user: User, appSettings: AppSettings): boolean {
        if (isSuperAdmin(user)) return true;
        // ...
    }
```

src/domain/usecases/SearchUsersAndUserGroupsUseCase.ts
```diff
@@ -0,0 +1,12 @@
+import { UseCase } from "../../CompositionRoot";
+import { FutureData } from "../entities/Future";
+import { UserSearch } from "../entities/UserSearch";
+import { UserSearchRepository } from "../repositories/UserSearchRepository";
+
+export class SearchUsersAndUserGroupsUseCase implements UseCase {
+    constructor(private userSearchRepository: UserSearchRepository) {}
+
+    public execute(query: string): FutureData<UserSearch> {
```
eperedo: Not a big deal, but since the search is only by name `filter: { displayName: { ilike: query } },` we should rename the variable from `query` to `name`.

src/domain/entities/Permission.ts
```diff
@@ -0,0 +1,7 @@
+import { NamedRef } from "./Ref";
+
+export type Permission = {
+    publicAccess: string; // '--------' | 'r-------'  | 'rw------'
```
eperedo: I think we don't need `publicAccess` since this functionality is only open to admin users + users/usersGroups that are included in the list (check my comment in the `src/domain/usecases/CheckCurrentUserCanAccessSettingsUseCase.ts` file).

If we're preparing this for the future we should not use the octal notation in the domain since that's a specific DHIS2 concept. We should use something like:

```ts
type AccessValue = {
  read: boolean;
  write: boolean;
  noAccess: boolean; // if AccessValue is a class this should be a calculated value using the `get` syntax.
 ```
 
 Then in the repository implementation convert to the octal notation: ` '--------' | 'r-------'  | 'rw------'`

eperedo (changes requested):
Thanks @p3rcypj, let me know if you have any questions.

src/domain/entities/Permission.ts
p3rcypj: I was thinking on the publicAccess also later for the actions because they should be public at the first glance... I'm going to think on that 👌🏼 because also we don't need the write permission actually, but we should have a way in case we need it on the future

p3rcypj: I've ended up removing `publicAccess` as you mentioned as this functionality is only open to admin users + users/userGroups, and a very created a very basic PublicPermissions to use it later for the actions. This way we specify in the domain which objects are the ones that could be accessible to everyone and we don't persist that information as it's a bussiness logic. So we end up with something like `settingsAccess: Permission`, but `actionsAccess: Record<..., PublicPermission>`. Do you like it?

eperedo (approved):

PR #310, opened 2025-02-25
Actions permissions

p3rcypj:
### :pushpin: References

-   **Issue:** Closes subtask https://app.clickup.com/t/8697qj21x (main task https://app.clickup.com/t/869656vnq)

### :memo: Implementation

Added one dropdown for each one of the actions for selecting which user groups has access to each action. If 'public access' is pressed it will remove the other selected user groups; and if there is no user group selected, it will select automatically 'public access'. Changes take place at the moment without reload. But admin users will always have all actions available even though permissions are set. Setting all actions available for users will reset all actions to public again.

image

https://github.com/user-attachments/assets/7a0e1319-8895-4fcb-874a-c01e4a59ece9


src/domain/usecases/CheckActionsAccessibleToCurrentUserUseCase.ts
```diff
@@ -0,0 +1,27 @@
+import _ from "lodash";
+import { ActionsPermissions } from "../entities/AppSettings";
+import { isSuperAdmin, User } from "../entities/User";
+import { getId } from "../entities/Ref";
+import { UserAction } from "../entities/UserAction";
+
+// Note for Reviewer: I made it sync to make it reactive, so I have the doubt if it shouldn't be a use case and just on the useHook
+// but as there is a CheckCurrentUserCanAccessSettingsUseCase it makes sense to have  this also as a usecase
+export class CheckActionsAccessibleToCurrentUserUseCase {
```
eperedo, line 9: I think this could be a hook since we already have all the information in the UI layer.

eperedo (changes requested):

PR #313, opened 2025-03-26
[bugFix]: Exclude openID, ldapID and externalAuth on user duplication

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8694phv1c

### :memo: Implementation
- [x] Remove openID, ldapID and external authentication details from user credentials on user duplication
  **Error message from import summary**:
  `"Property OIDC mapping value already exists, was given test_open_id"`
  **Fix:** 
  Each user in DHIS2 must have a unique OIDC mapping value if they are using OpenID Connect for login. So we exclude 
  OIDC values on user duplication. 

https://github.com/user-attachments/assets/a46fa356-6a58-45e7-bf8a-2ea4247441cc
https://github.com/user-attachments/assets/449a060e-b9d7-43d5-bb8d-60da9b72c765

### :fire: Testing

#8694phv1c


ifoche:
Task linked: [CU-8694phv1c Duplicating a user with OpenID causes errors](https://app.clickup.com/t/8694phv1c)

eperedo (changes requested):
I know we detected this bug in the replicate functionality (which is still part of the legacy code), but this would be a good opportunity to implement validation in the use cases: `ImportUsersUseCase` and `SaveUsersUseCase`. Both methods could potentially create multiple users with duplicate OpenID values.

Since we already have a `UserLogic.ts` entity, you could implement the validation logic there and reuse it in both use cases.

```ts
export class UserLogic extends Struct<User>() {
  static validateUniqueOpenId(users: User[]): boolean {
     // logic here
  }
}

// usecase

public execute({ users }: ImportUsersUseCaseOptions): FutureData<void> {
  if (!UserLogic.validateUniqueOpenId(users)) return Future.error(i18n.t("Open ID must be unique"));
        // ...
}
```

Here's a CSV file containing duplicate OpenID values you can use for testing the import process: 

[users-2025-03-26_17-15-06.csv](https://github.com/user-attachments/files/19476193/users-2025-03-26_17-15-06.csv)

src/domain/usecases/ImportUsersUseCase.ts
```diff
@@ -32,10 +33,14 @@ export class ImportUsersUseCase implements UseCase {
...
     public execute({ users }: ImportUsersUseCaseOptions): FutureData<void> {
         const usernameList = users.map(user => user.username);
-        return Future.join2(
-            this.userRepository.listAll({ filters: { "userCredentials.username": ["in", usernameList] } }),
-            this.userRepository.getCurrent()
-        ).flatMap(([usersFromDB, currentUser]: [User[], User]) => {
+        return Future.joinObj({
+            usersFromDB: this.userRepository.listAll({ filters: { "userCredentials.username": ["in", usernameList] } }),
+            currentUser: this.userRepository.getCurrent(),
+            allUsers: this.userRepository.listAll({}),
```
eperedo: Maybe we could improve peformance by only fetching users with an `openId` value, instead of pulling all users. Sounds good?

Something like this?

```bash
/api/users?filter=openId:!null
# or in case we don't have support for null in d2-api
/api/users?filter=openId:gt:0
```

src/domain/entities/UserLogic.ts
```diff
@@ -14,4 +14,10 @@ export class UserLogic extends Struct<User>() {
             user => user.organisationUnits.length > 0 && user.userRoles.length > 0 && user.userGroups.length > 0
         );
     }
+
+    static validateUniqueOpenId(users: User[], allUsers: User[]): boolean {
+        const openIds = users.map(user => user.openId);
+        const allOpenIds = allUsers.map(user => user.openId);
+        return openIds.every(openId => !allOpenIds.includes(openId));
```
eperedo: This works, but since this is a generic app we could have a server with thousands of users. Let's avoid having a nested loop (every -> includes). We can improve the validation by using a set:

```ts
static validateUniqueOpenId(users: User[], allUsers: User[]): boolean {
  const allOpenIdsSet = new Set(allUsers.map(user => user.openId));
  return users.every(user => !allOpenIdsSet.has(user.openId));
}
```

eperedo (changes requested):
Let me know if you have any questions

eperedo (approved):
Looks good, thanks @deeonwuli !

idelcano:
As this branch was part of https://github.com/EyeSeeTea/user-extended-app/pull/317
I already did the functional test, @MiquelAdell  

Since this branch was part of [this PR](https://github.com/EyeSeeTea/user-extended-app/pull/317),
I've already done the functional testing.
@MiquelAdell, from a functional point of view, I consider it approved. Unless you think we should hold off on merging or want to run any additional tests?
Thanks @deeonwuli

MiquelAdell:
Great @idelcano thank you all, merging it

PR #316, opened 2025-04-08
[fix]: Importing 500 users fails

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8694phpvk

### :memo: Implementation
- [x] Import users in chunks of 100 

https://github.com/user-attachments/assets/4449bf09-924a-4c10-8721-3cfca4fc21dc

### :fire: Testing
CSV file with 500 users for testing:
[import_bug.csv](https://github.com/user-attachments/files/19650934/import_bug.csv)

#8694phpvk


ifoche:
Task linked: [CU-8694phpvk Importing 500 users fails](https://app.clickup.com/t/8694phpvk)

src/domain/usecases/ImportUsersUseCase.ts
```diff
@@ -31,21 +31,24 @@ export class ImportUsersUseCase implements UseCase {
...
-            if (!hasRequiredFields)
-                return Future.error("All users must have at least one Organisation Unit, Role and Group");
+        return Future.futureMap(_.chunk(users, chunkSize), userChunk => {
+            const usernameList = userChunk.map(user => user.username);
 
-            const hasDuplicatedUsernames = _.uniq(usernameList).length !== usernameList.length;
-            if (hasDuplicatedUsernames) return Future.error("Usernames must be unique");
+            return Future.join2(
+                this.userRepository.listAll({ filters: { "userCredentials.username": ["in", usernameList] } }),
+                this.userRepository.getCurrent()
```
eperedo: Getting the current user every time is not necessary. We should fetch just once outside the loop.
```ts
return this.userRepository.getCurrent().flatMap((currentUser) => {
  return Future.futureMap(_.chunk(users, chunkSize), userChunk => {
   ...
  }
})
```

src/domain/usecases/ImportUsersUseCase.ts
```diff
@@ -31,21 +31,24 @@ export class ImportUsersUseCase implements UseCase {
...
     public execute({ users }: ImportUsersUseCaseOptions): FutureData<void> {
-        const usernameList = users.map(user => user.username);
-        return Future.join2(
-            this.userRepository.listAll({ filters: { "userCredentials.username": ["in", usernameList] } }),
-            this.userRepository.getCurrent()
-        ).flatMap(([usersFromDB, currentUser]: [User[], User]) => {
-            const hasRequiredFields = UserLogic.validateHasRequiredFields(users);
-            if (!hasRequiredFields)
-                return Future.error("All users must have at least one Organisation Unit, Role and Group");
+        return Future.futureMap(_.chunk(users, chunkSize), userChunk => {
```
eperedo: Nothing wrong with `futureMap` but I think in this context we should do sequential requests [instead parallel requests](https://github.com/EyeSeeTea/user-extended-app/blob/development/src/domain/entities/Future.ts#L101). This way it's easier to debug errors and we are sure next requests are not executed.

Something like this:

```ts
 return Future.sequential(_.chunk(users, chunkSize).map((usersChunk) => {
   ... 
 })
).toVoid();
```

eperedo (changes requested):
Thanks @deeonwuli. Just a few observations. Don’t hesitate to contact me with any questions.

eperedo (approved):

ifoche (changes requested):
Hey @deeonwuli . I have passed this to codex, and the suggestions it provides seem to be very meaningful. Could you please check them and address them?

Thank you!

- Duplicate detection only happens inside each chunk; duplicates spread across chunks slip through unchecked, risking conflicts or partial failures later.

Suggested task
Enforce global unique usernames during CSV import


- mergeUsers uses _.find to look up users in a map keyed by username, which needlessly scans all entries and obscures intent.

Suggested task
Use direct map lookup in mergeUsers


- Chunk-size values are hard-coded in multiple use cases (e.g., 50 vs 100), tying domain logic to transport limits and duplicating magic numbers.

Suggested task
Centralize and encapsulate chunking logic

ifoche (changes requested):
Sorry @deeonwuli but it seems like there's some error when trying to import from CSV here. I've tried to import a csv file with 510 users, to test the use-case that motivated the PR and I'm getting a weird error in a call to userGroups endpoint. Could you please check it?

NOTE: with the agreement of @anagperal I've also included some suggestions from codex, so you'll have to update your code before looking into it

Image

anagperal:
> Sorry @deeonwuli but it seems like there's some error when trying to import from CSV here. I've tried to import a csv file with 510 users, to test the use-case that motivated the PR and I'm getting a weird error in a call to userGroups endpoint. Could you please check it?
> 
> NOTE: with the agreement of @anagperal I've also included some suggestions from codex, so you'll have to update your code before looking into it
> 
> Image

@ifoche in this commit https://github.com/EyeSeeTea/user-extended-app/pull/316/commits/907959744eccd944a8da7d136978b05c9dca7f9b you have removed utils/chunks but also utils/limits is not added

ifoche (changes requested):
Hi @deeonwuli  

I just tired to replicate a user and got this SyntaxError. Could you please give it a try?

2025-10-02-130033_hyprshot

Could you please check it? Thank you!

ifoche (approved):
thanks @deeonwuli

PR #312, opened 2025-02-27
add new settings permissions page and new UI for roles, groups and dashboards

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869656vnq

### :video_camera: Screenshots/Screen capture

**List users, roles, groups and dashboards**

image

**Columns visibility** 

[User-Extended-App_columns_visibility.webm](https://github.com/user-attachments/assets/45274e50-d548-4b42-8847-0b96d868c0f1)

**Permissions settings**

image

**Hide users, roles or groups**

image

image

**User actions configuration**

image

**User Groups List**

image

**User Roles List**

image

**Dashboards**

image

### :fire: Testing

You can save the new settings in dataStore or a constant by setting the env. variable `REACT_APP_STORAGE=dataStore` or `REACT_APP_STORAGE=constants`

#869656vnq

PR #320, opened 2025-08-03
[Translations] Display UI with expected languages

p3rcypj:
### :pushpin: References

Closes https://app.clickup.com/t/8699wyrr0

### :memo: Implementation

**Issue**: _Interfaces not translated into French when the logged-in user is French._

Troubleshooting and issues encountered:
- French translations were partially available in the deprecated old translation files. Only 14 out of 151 strings were actually translated into French, making it impossible to display a decently translated French UI.
- New strings from `d2-i18n-extract` were not translated into French or Arabic, as only Spanish (`es.po`) was found in the `./i18n/` folder.
- Only 38 out of 116 strings were translated in `i18n/es.po`.
- Severe issues with multiple i18n instances and namespaces (d2 'translations' → later 'default', `d2-ui-components`, and `feedback-component`) completely broke translations.
- `nsSeparator: false` make TS complain by `i18next` (later added) types, while `nsSeparator: undefined` caused translation failures when running `yarn localize`.
- Translations from `d2-ui-components` were not being applied.
- `material-ui` components (also from `d2-ui-components`) were not being translated.
- Module declarations from `d2/lib/d2`, along with i18n types, were disallowing a fluent workflow.

Actions taken:
- Imported i18n from locales across all parts of the app. Done in bulk using regex.
- Ensured i18n is initialized wherever used (translations must be calculated with a function if declared outside components, to be invoked only after i18n is initialized). Manually reviewed usage file by file (regex was not reliable).
- Translated old translation files using GPT 4.1: `i18n_module_ar.properties`, `i18n_module_es.properties`, `i18n_module_fr.properties`.
- Spell-checked old translations using GPT 4.1.
- Added `fr.po` and `ar.po` → translated all new i18n strings for French and Arabic using GPT 4.1.
- Spell-checked new `fr.po` and `ar.po` translations with GPT 4.1.
- Spell-checked existing `es.po` lines (38 modified) with GPT 4.1.
- Added 78 new translations to `es.po` using GPT 4.1.
- Updated `d2-i18n`, `feedback-component`, and `d2-ui-components` (after several hard-to-resolve module and type errors).
  - While troubleshooting and updating dependencies i18n  types were linked and added alongside legacy `d2.i18n` types.
- Set `nsSeparator` to custom `':ns:'` instead of `false` or `undefined` (there was no better alternative to preserve `:` within strings).
- Inferred `material-ui` localization based on the user's `keyUiLocale` with country awareness (as `pt_PT` vs `pt_BR`).
- Added `d2/lib/d2` module declaration.
- Added portuguese (PT and BR) for both legacy i18n and the modern i18n one. (Translated and spell checked with GPT 4.1 and legacy ones also with LanguageTool.org)

_Note: All translations and spell checks with GPT 4.1 were performed with UI context awareness. Some additional minor suggestions were discarded due to lack of broader context._

Additional fixes:
- From legacy `.js`, some new key filter options used for export (added in #312) were chained from a `Maybe<boolean>` instance of filters, while in `.ts` they were declared as mandatory. As no linting was enforced on `.js`, and I had not tested the _Export_ functionality until reviewing translations, I discovered that export in other languages caused a runtime error and made it impossible to export any rows. Fix: [Chain `hideUsers` filter from `appContext`](https://github.com/EyeSeeTea/user-extended-app/pull/320/files#diff-60636b8ce7914c262054a19a98f2703f71977a1885ffd9d7f6165fc0c80973e1R651)
- Removed Spanish fuzzy

### Notes for reviewer :eyes:

For exports, column names are called via i18n. However, since exports are done to CSV or JSON, is i18n really necessary? Translations weren’t being applied anyway due to two reasons: language could not be changed, and i18n was called before initialization. Also, not all export columns exist in import columns, meaning we have two sources of truth — this could lead to unlikely but possible future issues. In my opinion, we should remove the use of i18n from `ExportUsersUseCase.ts`, and extend the columns after importing the mandatory ones from `ImportUsersUseCase.ts`. This could be treated as an exception to importing from another use case, or we could move `columnNameFromPropertyMapping` to `src/domain/usecases/common/utils.ts`.

[Added `ExportUsersUseCase.ts` FIXME comment](https://github.com/EyeSeeTea/user-extended-app/pull/320/files?new_files_changed=true#diff-843f953a2a4b447bb9cc15d19a1779a37bd9b36115ed0ea19ef00c011762ef99R14)  
[Reference for column mappings in `ImportUsersUseCase.ts`](https://github.com/EyeSeeTea/user-extended-app/blob/feature/fix-translations/src/domain/usecases/ImportUsersUseCase.ts#L9)

### :video_camera: Screenshots / Screen capture

image

### :fire: Testing

Languages available: Arabic, French, Spanish, and of course, English.

### :notebook: Dependencies upgraded

- Updated `d2-i18n` to v1.1.3 (adds type declarations; already used in `d2-ui-components`)
- Updated `d2-ui-components` to v2.11.0
- Updated `feedback-component` to v0.3.0


src/utils/i18n.ts
```diff
@@ -0,0 +1,6 @@
+import i18n from "../locales";
+
+const clone = i18n.cloneInstance();
+clone.setDefaultNamespace("user-extended-app");
```
eperedo: We already have an implementation to avoid problem with translations. Can you check [this PR](https://github.com/EyeSeeTea/user-extended-app/pull/315) and keep just one implementation to avoid duplicate code or merge errors. Also I think this fix the nsSeparator problem.

eperedo (changes requested):

src/utils/i18n.ts
p3rcypj: Thanks a lot, @eperedo, for your review! I've tried to take a _middle 50–50_ approach in the end, keeping the changes from #315 while also preserving the i18next inferred types. Please have a look and see if you think it's a good change. If so, I should also reflect these new changes on all of our repositories as @gqcorneby did.

Also, @tokland, since you added the types in [Improve typed i18n infrastructure + explicit namespace](https://github.com/EyeSeeTea/dhis2-app-skeleton/commit/54fcc221503d23f3bf4490d4315679b4cf60034b) on the **dhis2-app-skeleton** repository, could you check the change I’ve made to this file? Now, with the _@dhis2/d2-i18n_ package upgrade, we can infer i18next types as it now includes i18n types (inferred as well).  

P.S.: The `cloneInstance` here was just a quick fix to be on time for last week’s meeting.  

What actually worries me is that in `locales/index.js`, `setDefaultNamespace` is being used, which causes issues. My concern is that it could override the “default” namespace —literally named “default”— on DHIS2 `t` function. If DHIS2 somehow reloads state, its i18n translation function may not work as expected, because it would look in our namespace instead. By the way for debugging this: it's now accessible via `i18n.options.defaultNS`.

gqcorneby: Just adding this note. There's a way to generate the locales in "lib mode" which removes `setDefaultNamespace` and `changeLanguage` in the generated `locales/index.js`. Links below for reference. 

https://github.com/EyeSeeTea/feedback-component/pull/7
https://github.com/EyeSeeTea/feedback-component/pull/7#issuecomment-2771992256

src/utils/i18n-typed.ts
```diff
@@ -0,0 +1,17 @@
...
+    const { t: _t, changeLanguage, ...rest } = i18n;
+
+    const restBinded = Object.fromEntries(
+        Object.entries(rest).map(([key, value]) => [key, typeof value === "function" ? value.bind(i18n) : value])
+    ) as typeof rest;
+
+    return {
+        ...restBinded,
+        t: function (s: string | string[], options?: i18n.TranslationOptions<object>): string {
+            return i18n.t(s, { ...options, ns: namespace, nsSeparator: ":ns:" });
```
eperedo: Although this works, I think we should keep using `false` for separators until we decide to implement this in the skeleton app, to avoid having different implementations across our apps.

From what I’ve seen, newer versions still keep the boolean as an alternative, so it’s probably the safest choice for now.

 
 https://www.npmjs.com/package/@types/i18next/v/11.9.3?activeTab=code
 https://github.com/i18next/i18next/blob/a7852d7698ec11ccd83b8da914c1eb859b4460dd/typescript/options.d.ts#L601

eperedo (changes requested):

eperedo (approved):

PR #321, opened 2025-08-11
User metadata visualizations UI refinements

p3rcypj:
### :pushpin: References

-   **Issue:** Closes #?

### :memo: Implementation

- Reorganise setting 'show only users assigned to users' organisation units' (only users tab for now)
- Add set password action
- Reorganize action permissions with rules

### :fire: Testing

[*WIP*]

📋 Test Coverage Created
29 test cases covering all validation scenarios
6 main test categories:
Empty password validation - handles null, undefined, and empty strings
Length validation - minimum 8 characters, maximum 34 characters
Lowercase letter validation - requires at least one lowercase letter
Uppercase letter validation - requires at least one uppercase letter
Digit validation - requires at least one number
Special character validation - requires at least one special character
🔍 Test Categories
Empty Password Tests (3 tests)

Empty string, undefined, and null inputs
Length Validation (5 tests)

Passwords too short (<8 chars)
Passwords too long (>34 chars)
Boundary conditions (exactly 8 and 34 chars)
Character Requirements (8 tests)

Each character type requirement tested individually
Both positive and negative test cases
Special Character Coverage (4 tests)

Common special characters (!@#$%^&* etc.)
Unicode special characters (€¡¿§)
Whitespace as special characters
Comprehensive Scenarios (4 tests)

Valid passwords meeting all requirements
Multiple validation failures
Validation order verification
Edge Cases & Boundaries (5 tests)

Mixed character sets
Unicode handling
Repeated characters
Whitespace variations
✅ Test Results
All 29 tests passing ✓
100% validation logic coverage
DHIS2-compliant validation rules verified

Summary
I've successfully created comprehensive tests for all the requested functions in [User.ts](vscode-file://vscode-app/c:/Users/pablo/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html). Here's what was accomplished:

Functions Tested:
✅ [isSuperAdmin](vscode-file://vscode-app/c:/Users/pablo/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) - Tests for users with/without "ALL" authority, including edge cases
✅ [hasReplicateAuthority](vscode-file://vscode-app/c:/Users/pablo/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) - Tests for users with/without "F_REPLICATE_USER" authority
✅ [allUsersHaveAllSpecifiedAccesses](vscode-file://vscode-app/c:/Users/pablo/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) - Tests access validation with various scenarios
✅ [userWithinOrgUnits](vscode-file://vscode-app/c:/Users/pablo/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) - Tests org unit membership including inheritance (exported for testing)
✅ [userInheritedOrgUnitIds](vscode-file://vscode-app/c:/Users/pablo/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) - Tests inheritance chain logic (exported for testing)
✅ [userOrgUnitIds](vscode-file://vscode-app/c:/Users/pablo/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) - Tests basic org unit ID extraction
✅ [allUsersHaveEmail](vscode-file://vscode-app/c:/Users/pablo/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) - Tests email validation for arrays of users
✅ [allUsersBelongToAtLeastOneOrgUnit](vscode-file://vscode-app/c:/Users/pablo/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) - Tests org unit membership for user collections
✅ [allUsersAreDisabled](vscode-file://vscode-app/c:/Users/pablo/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) - Tests disabled status validation
✅ [allUsersAreActive](vscode-file://vscode-app/c:/Users/pablo/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) - Tests active status validation
Test Fixtures Created:
Super admin user with "ALL" authority
Replicate user with "F_REPLICATE_USER" authority
Regular user with basic permissions
User without email
Disabled user
User with multiple org units
Hierarchical org unit structure (root → child → grandchild)
Various access permission combinations
Key Test Cases Covered:
Edge cases: Empty arrays, undefined/null values, empty strings
Boundary conditions: Users with no org units, no authorities, no email
Hierarchical inheritance: Testing org unit path inheritance logic
Permission combinations: Various access permission scenarios
Status validation: Active vs disabled users
Email validation: Including current behavior with whitespace
Code Improvements Made:
Made [userWithinOrgUnits](vscode-file://vscode-app/c:/Users/pablo/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [userInheritedOrgUnitIds](vscode-file://vscode-app/c:/Users/pablo/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) public for better testability
Enhanced [isSuperAdmin](vscode-file://vscode-app/c:/Users/pablo/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [hasReplicateAuthority](vscode-file://vscode-app/c:/Users/pablo/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) functions to handle undefined authorities gracefully using optional chaining
Test Results:
✅ 45 tests passing
✅ 0 failures
✅ Project builds successfully
✅ No TypeScript errors
The test suite provides comprehensive coverage of all the requested functions with realistic test data that matches the existing codebase patterns. The tests validate both the happy path scenarios and edge cases, ensuring robust validation of the User entity functions.


src/domain/entities/ActionPermission.ts
```diff
@@ -0,0 +1,51 @@
...
+            users: [],
+            userGroups: [],
+            rules: [],
+        });
+    }
+
+    get isPublic(): boolean {
+        return _.isEmpty(this.users) && _.isEmpty(this.userGroups) && _.isEmpty(this.rules);
+    }
+    // get isPublic(): boolean {
```
eperedo: can we delete this comment? If we need for some reason maybe adding more context about it would be great.

src/domain/entities/ActionPermission.ts
```diff
@@ -0,0 +1,51 @@
...
+import { Id, NamedRef } from "./Ref";
+import { Rule } from "./Rule";
+
+type ActionPermissionAttrs = PermissionAttrs & {
+    rules: Rule[];
+};
+
+export class ActionPermission extends Struct<ActionPermissionAttrs>() {
+    static public(): ActionPermission {
+        return new ActionPermission({
```
eperedo: Since we're using Struct this could be `this.create(...)`, right?

src/domain/entities/AppSettings.ts
```diff
@@ -61,10 +62,59 @@ export class AppSettings extends Struct<AppSettingsAttr>() {
...
+
+    private static defaultColumns(): SettingsUserColumn[] {
+        const defaultColumns = getDefaultUserColumns();
+        return defaultColumns.map(column => ({ field: column.name, value: "optional" }));
+    }
+}
+
+export function markAllActionsPublic() {
+    const publicPermission = ActionPermission.public();
+    const publicActions = Object.assign({}, ...userActions.map(action => ({ [action]: publicPermission })));
```
eperedo: can we add the type here?

```ts
    const publicActions: ActionsPermissions = Object.assign({}, ...userActions.map(action => ({ [action]: publicPermission })));
```

Otherwise we get publicActions as any. I think we should use satisfies here, but not supported in this TS version.

src/domain/entities/User.ts
```diff
@@ -92,26 +92,54 @@ export interface AccessPermissions {
...
 }
 
+// FIXME: We need tests for these functions
 export const isSuperAdmin = (user: User): boolean => {
     return user.authorities.includes("ALL");
 };
 
 export const hasReplicateAuthority = (user: User): boolean => {
-    return _.some(user.authorities, authorities => authorities.includes("F_REPLICATE_USER"));
+    return user.authorities.includes("F_REPLICATE_USER");
```
eperedo, line 102: [no need to change anything here] 
I know this is something we introduced a long time ago so just as a reference for the future:

I think "authorities" should not be in the domain since this is a DHIS2 concept and we only need the values to calculate user permissions. 

we should have:
```ts
type User = {
  ...,
  canReplicateUser: boolean;
}
```

and then we generate the value in the implementation `UserD2ApiRepository`.

src/webapp/components/permissions-page/PermissionsPage.tsx
```diff
@@ -56,23 +56,29 @@ export const PermissionsPage = React.memo((props: PermissionsPageProps) => {
...
+                            onChange={event => updateFormState(event.target.checked, "showHideOrgUnits")}
                         />
                     }
-                    label={i18n.t("Show only active users")}
+                    label={i18n.t("Hide org units on advanced filters")}
                 />
 
+                {formState.showHideOrgUnits && (
+                    <Box marginTop={1} marginBottom={3}>
+                        Placeholder for org units tree
```
eperedo, line 79: Is this the correct text? Also we should have i18n here, right?

src/webapp/components/users-selected-modal/PasswordRequirements.tsx
```diff
@@ -0,0 +1,57 @@
...
+                        {requirement.text()}
+                    </RequirementItem>
+                ))}
+            </RequirementsList>
+        </RequirementsContainer>
+    );
+});
+
+const RequirementsContainer = styled.div`
+    margin-top: 8px;
```
eperedo: can we use logical properties for this? 

```css
margin-block-start: 8px;
```

https://github.com/EyeSeeTea/know-how/wiki/UI-Guide#using-css-logical-properties

Also check other css code introduced in this PR

eperedo (changes requested):
PR is not ready, but Pablo requested a code review.

src/data/repositories/InstanceD2ApiRepository.ts
```diff
@@ -36,4 +38,26 @@ export class InstanceD2ApiRepository implements InstanceRepository {
             )
         );
     }
+
+    verifyPassword(password: string): FutureData<true> {
```
eperedo: I think this should be in the `UserRepository` and not here since is not related to an Instance.

A `PasswordValidation` entity also could be a good idea in case we think this could be something not always related to a user.

src/domain/usecases/CheckActionsAccessibleToCurrentUserUseCase.ts
```diff
@@ -1,27 +1,91 @@
...
+    userOrgUnitIds,
+    allUsersBelongToAtLeastOneOrgUnit,
+} from "../entities/User";
+import { getId, Id } from "../entities/Ref";
 import { UserAction } from "../entities/UserAction";
+import { UserActionRule } from "../entities/UserActionRule";
+import { ActionPermission } from "../entities/ActionPermission";
 
-// Note for Reviewer: I made it sync to make it reactive, so I have the doubt if it shouldn't be a use case and just on the useHook
+// Old Note for Reviewer: I made it sync to make it reactive, so I have the doubt if it shouldn't be a use case and just on the useHook
```
eperedo: Yes, this could live in a hook for now since we don't need any interaction with repositories.

src/domain/usecases/SetUserPasswordUseCase.ts
```diff
@@ -0,0 +1,12 @@
+import { MetadataResponse } from "@eyeseetea/d2-api/2.36";
+import { FutureData } from "../entities/Future";
+import { User } from "../entities/User";
+import { UserRepository } from "../repositories/UserRepository";
+
+export class SetUserPasswordUseCase {
+    constructor(private userRepository: UserRepository) {}
+
+    public execute(userToSave: User): FutureData<MetadataResponse> {
+        return this.userRepository.save([userToSave]);
```
eperedo: I think we can reuse the `SaveUsersUseCase`, right? No need to create another one since we don't need specific logic for this process.

src/domain/usecases/VerifyPasswordUseCase.ts
```diff
@@ -0,0 +1,10 @@
+import { FutureData } from "../entities/Future";
+import { InstanceRepository } from "../repositories/InstanceRepository";
+
+export class VerifyPasswordUseCase {
+    constructor(private instanceRepository: InstanceRepository) {}
+
+    public execute(password: string): FutureData<true> {
+        return this.instanceRepository.verifyPassword(password);
```
eperedo: Check the comment in the `InstanceD2ApiRepository.ts` file

src/webapp/components/users-selected-modal/PasswordRequirements.tsx
```diff
@@ -0,0 +1,57 @@
+import React from "react";
+import styled from "styled-components";
+import { passwordRequirements } from "./passwordValidation";
+import i18n from "../../../locales";
+
+interface PasswordRequirementsProps {
+    password: string;
+}
+
+export const PasswordRequirements: React.FC<PasswordRequirementsProps> = React.memo(({ password }) => {
```
eperedo, line 10: Not sure if I'm missing something here, but this component is not being used, right?

src/webapp/components/users-selected-modal/__tests__/validatePasswordRules.spec.ts
```diff
@@ -0,0 +1,326 @@
...
+            expect(result).toBe("Password is required");
+        });
+
+        it("should return error for null", () => {
+            const result = validatePasswordRules(null as any);
+            expect(result).toBe("Password is required");
+        });
+    });
+
+    /* Commented for reference
```
eperedo: I think we can replace these comments with the link to the docs for further details about these rules, what do you think?

https://docs.dhis2.org/en/use/user-guides/dhis-core-version-241/configuring-the-system/users-roles-and-groups.html#password-requirements

eperedo (changes requested):

src/domain/entities/User.ts
p3rcypj: @eperedo I completely agree with you. Good point! I was just changing this old code from Jocelyn but as you point out, is something that should be filled in `UserD2ApiRepository`. Or even being User a class and `canReplicateUser` a private property only accessible by `get()`

src/webapp/components/permissions-page/PermissionsPage.tsx
p3rcypj: @eperedo It's actually not something for this PR, at first I started this PR as refinements-ui because my plan was to set a lot of UI tasks and later implement the functionality. But after some weeks and considering the time it was taking me to handle the tasks, we focused on this PR only be for the "Set password" and moving the permission filter to advanced filters.

src/webapp/components/users-selected-modal/PasswordRequirements.tsx
p3rcypj: It's not being used, no. At first the validation was done on the client side. But later I implemented validation through DHIS2. I let this code because I'm not sure which way should be better.

src/data/repositories/InstanceD2ApiRepository.ts
p3rcypj: I was thinking the same. To be honest I was unsure because verifying the password it's in DHIS2 directly related to the instance settings and I was unsure if it was used somewhere else. Now I checked on the API docs and it seems only applies to User's. So as you also see it that way, I'll move it because it makes more sense even if for DHIS2 it's explicitly related on the origin.

src/domain/usecases/SetUserPasswordUseCase.ts
p3rcypj: Actually, at first glance, I was actually going to say "Oh that's a good point, I didn't notice", because I think it is since we don't need specific logic. But now that I think it twice, we don't do usecases just because we need some specific code, but because it's an action we want to perform. In this case for example, we are forcing on the domain, that we only want one user at a time to set the password. Also, now that I'm typing this comment, maybe we should have the DHIS2 validation here on the usecase before saved, and use the client side validation on the presentation... what do you think? @eperedo

src/webapp/components/users-selected-modal/__tests__/validatePasswordRules.spec.ts
p3rcypj: Indeed, good point. Also anyways, it's more about the decision of: whether we should validate on the client side meaning if DHIS2 changes someday the requirements we will have to maintain them, or if we validate on server-side. @eperedo

src/domain/usecases/SetUserPasswordUseCase.ts
eperedo: @p3rcypj Good point, let's keep this usecase since we have specific logic for this process. And yes including password validation here is also necessary.

src/webapp/components/permissions-page/PermissionsPage.tsx
eperedo, line 79: Then we can delete this part, right?

eperedo (changes requested):
[code review] Looks good, just one minor observation. Please check my comment.

eperedo (approved):

PR #327, opened 2025-09-23
ignore blacklisting for admin users

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869a4q1n9

### :memo: Implementation

- [x] apply hide users, userGroups and userRoles in all pages: users, userRoles, userGroups and dashboards.
- [x] ignore configuration if current user is admin

### :fire: Testing

#869a4q1n9

PR #329, opened 2025-10-02
Not forcing root junction when 'Show only active users'

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869a4nb2c

### :memo: Implementation

This PR depends on https://github.com/EyeSeeTea/user-extended-app/pull/327

- [x] Administrators can modify the "Active users" filter
- [x] OR/AND filter is always enabled. 
- [x] We simulate the `OR` filter by making multiple requests (base filters + other filters), removing duplicates, and handling pagination on the client side

### :fire: Testing

#869a4nb2c


ifoche:
Task linked: [CU-869a4nb2c Not forcing root junction when 'Show only active users' and 'Show users within my org unit'](https://app.clickup.com/t/869a4nb2c)

PR #333, opened 2025-10-13
Refactor: improve replicate performance

nshandra:
### :pushpin: References

-   **Issue:** Closes [Improve performance on replicate action](https://app.clickup.com/t/869af8h25)

### :memo: Implementation

Improve replicate forms loading time by retrieving only the Id and Username of the instance users instead of the full user data. Originally, this resulted in a lengthy wait in instances with a large amount of users with many Roles and Groups assigned for each of them.


xurxodev (approved):
Thanks @nshandra 

The performance to load of dialog to import from table improve a lot with this change.

The task description says about poor performance to save also. Yo told me that you don't see possible to improve this part.

@adrianq I don't know if you want that I review the save to analize it with other eyes.

@nshandra Only a minor comment:

- Try to avoid to use useSnackbar in hooks like useGetAllUserIdentifiers.

The reason is that mix render a presentation logic. It's better to return an error and in the component decide to use snackbar or other mechanism to render the error.

adrianq:
@xurxodev yes, discuss with @nshandra  if there is a not-too-difficult solution for the save.

nshandra:
@xurxodev @adrianq I didn't included it because i mentioned it to both of you in meets, but changing the save to POSTs to `/api/users` or `api/users/{Id}/replicate` was significantly slower in a instance with many users, roles and groups.

Tests results:
Sierra leone DB:
Replicate API 10 users: 1.3 s
Current method 10 users: 2 s

EST DEV like instance:
Replicate 1 user via native User app: ~40 s

Replicate API 10 users(sequential): ~260 s
Replicate API 10 users(batches of 5): ~140 s

Post to /users (sequential) 10 users: ~400 s
Post to /users (batches of 5) 10 users: ~320 s

Current method 10 users: 7,5 s

PR #334, opened 2025-11-10
fix: reduce concurrency to avoid deadlocks

MatiasArriola:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8699r8t7c

### :memo: Implementation

- Reduce api requests concurrency to avoid deadlocks on the server
  - This happened in specific environments only (dev)
  - These requests might end up locking each other (update userinfo locks + userinfo locks from FK update usersetting)
- The process will take more time but should be safer


anagperal (approved):
thanks @MatiasArriola !

ifoche (approved):
thanks @MatiasArriola

PR #339, opened 2026-03-18
fix: reload settings/table after updating settings

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869bq57vu

### :memo: Implementation
**Problem**: 
- Updating import settings did not refresh the user table or the settings used by import/export actions. As a result, changes like `organisationUnitsField` were only applied after a manual reload or remount.

**Solution**:
- [x] Added a reloadable settings flow in `SettingsDialogModal.tsx` and updated `UserListTable.tsx` to re-fetch import settings when the table reloads and immediately after settings are saved.

https://github.com/user-attachments/assets/c3cf99af-66af-48cf-996e-1b6a199a50e2

### :fire: Testing

#869bq57vu


ifoche:
Task linked: [CU-869bq57vu import/export org unit type selection setting only works on import](https://app.clickup.com/t/869bq57vu)

src/webapp/components/user-list-table/UserListTable.tsx
```diff
@@ -419,6 +419,10 @@ export const UserListTable: React.FC<UserListTableProps> = ({
 
     const tableProps = useObjectsTable(baseConfig, refreshRows, refreshAllIds);
 
+    useEffect(() => {
```
gqcorneby: [minor optimization] I think this useEffect creates a double-fetch of `Settings.build` on every `reloadTableKey` change, because the hook already re-runs its own internal effect when that key changes.

If we also want settings to reload when reloadKey changes (i.e. after reload() is called from onSettingsClose), I think it would be good to pass a combined key directly into the hook and remove this effect entirely:

const { importSettings, reloadImportSettings } = useImportSettings(`${reloadTableKey}-${reloadKey}`);

Then in onSettingsClose, remove the direct `reloadImportSettings()` call. `reload()` alone will change reloadKey, the hook will respond, and settings will reload once via a single code path.

gqcorneby (commented):
Thanks @deeonwuli! I just have one comment about a minor optimization that's worth considering :)

src/webapp/components/user-list-table/UserListTable.tsx
deeonwuli: Good catch. Thank you 👍

adrianq:
@deeonwuli A merge is needed here as there are soem conflicts

deeonwuli:
@gqcorneby After merging with development, the settings modal visibility was moved to `Router.tsx` and this broke the reload flow. To fix this, we are now passing the `reloadKey` from `Router.tsx` component down to `ListComponent.js` and finally to `UserListTable` where it is used in the `useImportSettings` hook.

gqcorneby (approved):
All good! Thanks @deeonwuli!

PR #343, opened 2026-04-15
Feature/changes user role user groups tabs

xurxodev:
### :pushpin: References

-   **Issue:** Closes  https://app.clickup.com/t/869c4afga #869c4afga

### :memo: Implementation                                                                                                                                                                            
                                                                                                                                                                                                       
  This branch consolidates several changes to the **User Groups**, **User Roles** and **Dashboards** tabs, both in behaviour and in App Settings.                                                      
                                                                                                                                                                                                       
  **1. Hide "Filter by Users" column in DHIS2 2.42** (`3ba477e`)
                                                                                                                                                                                                       
  DHIS2 2.42 moved several user properties out of `userCredentials`. The "Filter by Users" column and related filter is now hidden on 2.42 instances where the filtering contract is incompatible,
  avoiding broken queries.                                                                                                                                                                             
                                                            
  **2. Add Hide not applicable user roles filter** (`d57f97b`)                                                                                                                           
                                                            
  User Roles tab now mirrors User Groups behaviour: after filtering users by the current org unit scope, roles with no remaining users are hidden in memory when "Hide not applicable user roles" is   
  enabled. Previously this only worked via API for User Groups.

  **3. Shared implementation between User Roles and User Groups** (`c2890be`)                                                                                                                          
                                          
  Refactor to remove duplicated filtering / org-unit scoping logic between the two tabs, so they now share the same code paths.                                                                        
                                                                                                                                                                                                       
  **4. Load all users when org-unit scope is disabled in users filter** (`cb4c056`)
                                                                                                                                                                                                       
  When the "Show only users assigned to my organization unit and below" toggle is **off**, the "Filter by Users" dropdown now fetches **all** users via a new `GetAllUsersSimpleUseCase`, so the admin
  can filter by users outside the current scope.                                                                                                                                                       
                                                            
  **5. Configurable default value for boolean filters in App Settings** (`fef5a3d`)                                                                                                                    
   
  App Settings → **Filter Permissions** can now configure, per tab (User Groups / User Roles / Dashboards), not only **Visible** but also **Default** for the boolean filters:                         
  - `filterUsersInOrgUnit` (all three tabs)                 
  - `filterHideNotApplicableUserGroups` / `filterHideNotApplicableUserRoles`
                                                                                                                                                                                                       
  The `Show/Hide Filters` section is renamed to `Filters` and uses a two-column `Visible | Default` layout. `Visible` and `Default` are independent — the admin can pre-configure the default state
  even when the filter is hidden from non-admin users. Defaults apply to admin and non-admin alike.                                                                                                    
                                                                                                                                                                                                       
  On the data layer, `appSettingsHelpers.mergeAndAddRuntimeProps` was refactored into a single generic `migrateRecord` helper that back-fills missing keys and missing fields from defaults. This      
  replaces the previous `migrateNewerActions` and merge logic, and covers forward-compatible migration for both class-based (`actionsAccess`) and plain-object records (`uiXxxActionsAccess`). Unit    
  tests added in `src/data/repositories/common/__tests__/appSettingsHelpers.spec.ts`.                                                                                                                  
                                                                                                                                                                                                       
App settings Before
<img width="1900" height="930" alt="Screenshot 2026-04-15 at 10 19 41" src="https://github.com/user-attachments/assets/733151e4-7a69-47e5-882e-e52721050dd6" />


App settings After
<img width="1918" height="931" alt="Screenshot 2026-04-15 at 10 18 22" src="https://github.com/user-attachments/assets/a5480a53-8fa1-4064-989c-58d046aeac3f" />

                                              
  ### :fire: Testing                      
                                              
  1. Open the app as **admin** and go to **Settings → Filter Permissions**.                                                                                                                            
  2. Expand each accordion (User Groups, User Roles, Dashboards) and verify:
      - The section is titled **Filters** and shows two columns **Visible | Default**.                                                                                                                 
      - Boolean filters (`Show only users assigned...`, `Hide not applicable...`) have both switches.
      - Non-boolean rows (`Filter by Users`, `Filter by Owners`, exports) show only **Visible**, with `—` under Default.                                                                               
  3. Toggle `Show only users assigned to my org unit and below` **Default = OFF** in User Groups and save.                                                                                             
  4. Go to the **User Groups** tab and verify the filter opens already unchecked and the full user-group list (beyond the admin's org units) is displayed.                                             
  5. Repeat step 3–4 for **User Roles** and **Dashboards**.                                                                                                                                            
  6. Set `filterUsersInOrgUnit.visible = false` but `defaultValue = false` and save. Log in as a **non-admin** user and verify:                                                                        
      - The toggle is not visible in the User Groups / User Roles / Dashboards tab.                                                                                                                    
      - The list loads **without** the org-unit scope applied (hidden default is honoured).                                                                                                            
  7. With `Default = ON`, open the User Roles tab on a DHIS2 2.42 instance and verify:                                                                                                                 
      - Roles with no users from the current org unit scope are hidden when "Hide not applicable user roles" is active.                                                                                
      - Disabling the org-unit toggle loads the full user list in the "Filter by Users" dropdown.                                                                                                      
  8. Run `yarn test` — the new `appSettingsHelpers.spec.ts` suite (6 cases) must pass.                                                                                                                 
  9. Verify backward compatibility: on an instance with App Settings already stored **without** `defaultValue`, reload the app and confirm nothing breaks (missing `defaultValue` is back-filled to    
  `true` by `migrateRecord`).

### DIFF
--- a/i18n/ar.po
+++ b/i18n/ar.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: i18next-conv\n"
-"POT-Creation-Date: 2026-03-25T05:57:54.676Z\n"
+"POT-Creation-Date: 2026-04-15T08:09:38.204Z\n"
 "PO-Revision-Date: 2018-10-25T09:02:35.143Z\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
@@ -47,17 +47,20 @@ msgstr ""
 msgid "Show Export to CSV"
 msgstr ""
 
+msgid "Show only users assigned to my organization unit and below"
+msgstr ""
+
+msgid "Hide not applicable user groups"
+msgstr ""
+
+msgid "Filter by Users"
+msgstr ""
+
 #, fuzzy
-msgid "Show \"Show only users assigned to my organization unit and below\""
-msgstr "إظهار فقط المستخدمين المعينين لوحدات تنظيم المستخدمين"
+msgid "Hide not applicable user roles"
+msgstr "تكرار المستخدم من الجدول"
 
-msgid "Show \"Hide not applicable user groups\""
-msgstr ""
-
-msgid "Show \"Filter by Users\""
-msgstr ""
-
-msgid "Show \"Filter by Owners\""
+msgid "Filter by Owners"
 msgstr ""
 
 msgid "ID"
@@ -203,9 +206,6 @@ msgid "Owners: {{owners}}"
 msgstr ""
 
 msgid "Users: {{users}}"
-msgstr ""
-
-msgid "Show only users assigned to my organization unit and below"
 msgstr ""
 
 msgid "Filters"
@@ -386,6 +386,9 @@ msgstr ""
 
 msgid "Dashboards"
 msgstr "لوحات المعلومات"
+
+msgid "Value"
+msgstr ""
 
 #, fuzzy
 msgid "Select user groups and rules"
@@ -709,9 +712,6 @@ msgstr "تكرار المستخدم من القالب"
 msgid "Replicate user from table"
 msgstr "تكرار المستخدم من الجدول"
 
-msgid "Hide not applicable user groups"
-msgstr ""
-
 msgid "Loading..."
 msgstr ""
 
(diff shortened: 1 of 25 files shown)


adrianq:
@xurxodev all good - just a minor fix I have implemented here https://github.com/EyeSeeTea/user-extended-app/pull/343/commits/64627bdf720f2231193dd49c73e6cec62dd45db3 if you confirm it is OK with you I will merge the PR

xurxodev:
@adrianq Nice catch, I missed this one. The fix looks good to me.

PR #347, opened 2026-05-04
support dhis v42 header

adrianq:
### :pushpin: References

-   **Issue:** Closes #?


xurxodev (approved):
thanks @adrianq 

I've tested in 2.42 and works 

Image

The code is similar to other proyects like this PR:
https://github.com/EyeSeeTea/dhis_tally_sheets/pull/25

* Create HeaderBar with fallback 
* Use the new HeaderBar instaead of HeaderBar from @dhis2/ui

PR #350, opened 2026-06-02
fix: handle null userCredentials when accessing username in import an…

xurxodev:
### :pushpin: References

- **Issue:** Closes https://app.clickup.com/t/869dfnjjm

### :memo: Implementation

In some DHIS2 instances, users can exist without a `username` field (confirmed via `GET /api/users?filter=username:null`). When this happens, the code falls through to `user.userCredentials.username` — but in DHIS2 2.42+ `userCredentials` is no longer returned by the API, so this throws an error.

```typescript
  new UserIdentifier({  id: user.id, username: user.username ?? user.userCredentials.username, name: user.name,})
```

The fix applies optional chaining and a root-level `username` fallback consistently across all places that access `userCredentials.username`:

```typescript
new UserIdentifier({
    id: user.id,
    username: user.username || user.userCredentials?.username || "",
    name: user.name,})
```

> Note: the underlying cause is users in the instance with no username. Ideally those users should be fixed at the data level, but this change prevents the app from crashing when it encounters them.

### :fire: Testing

- List users: page loads without errors

### DIFF
--- a/src/data/repositories/UserD2ApiRepository.ts
+++ b/src/data/repositories/UserD2ApiRepository.ts
@@ -339,7 +339,7 @@ export class UserD2ApiRepository implements UserRepository {
                         user =>
                             new UserIdentifier({
                                 id: user.id,
-                                username: user.username ?? user.userCredentials.username,
+                                username: user.username || user.userCredentials?.username || "",
                                 name: user.name,
                             })
                     );
@@ -393,7 +393,7 @@ export class UserD2ApiRepository implements UserRepository {
                 user =>
                     new UserIdentifier({
                         id: user.id,
-                        username: user.username || user.userCredentials.username,
+                        username: user.username || user.userCredentials?.username ,
                         name: user.name,
                     })
             );
--- a/src/data/repositories/UserD2ApiRepository.ts
+++ b/src/data/repositories/UserD2ApiRepository.ts
@@ -393,7 +393,7 @@ export class UserD2ApiRepository implements UserRepository {
                 user =>
                     new UserIdentifier({
                         id: user.id,
-                        username: user.username || user.userCredentials?.username ,
+                        username: user.username || user.userCredentials?.username || "",
                         name: user.name,
                     })
             );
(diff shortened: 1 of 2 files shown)


src/data/repositories/UserD2ApiRepository.ts
```diff
@@ -393,7 +393,7 @@ export class UserD2ApiRepository implements UserRepository {
                 user =>
                     new UserIdentifier({
                         id: user.id,
-                        username: user.username || user.userCredentials.username,
+                        username: user.username || user.userCredentials?.username ,
```
tokland: Above we have `user.username || user.userCredentials?.username || ""`, do we also need here the `|| ""` fallback?

Side note: TS didn't catch it because `userCredentials` is inferred by d2-api as present, while we know it won't be for newer DHIS2 versions. However, I think it's ok to take no action about the type for now; once user-extended is updated to use d2-api@4x, we will get the correct typings automatically.

src/legacy/models/userHelpers.js
```diff
@@ -460,7 +460,7 @@ async function saveCopyInUsers(d2, users, copyUserGroups) {
             fields: ":owner,userCredentials,userGroups[id]",
             filter:
                 `${is242Plus ? "username" : "userCredentials.username"}:in:[` +
-                _(users).map("userCredentials.username").join(",") +
+                _(users).map(user => user.username || user.userCredentials?.username).join(",") +
```
tokland: If we have no username for some user, we may end with `["id1", undefined, "id2"].join(",") => 'id1,,id2'`. If that's a problem, we may compact the array before joining. And then we may need to check the filter length to be > 1 (which was not done in the existing code anyway)

(probably a non-relevant edge case, but just to mention it).

tokland (commented):
In-line comments:

src/data/repositories/UserD2ApiRepository.ts
xurxodev: Good catch, added the || "" fallback for consistency. Noted on the typings — agreed, no action needed until the d2-api@4x upgrade.

src/legacy/models/userHelpers.js
xurxodev: Good point. Added .compact() to filter out falsy usernames before joining, so we avoid the 'id1,,id2' case. Agreed on the filter length check — leaving that as a follow-up since it was pre-existing.

tokland (approved):

PR #353, opened 2026-06-15
[feature]: Allow non-admin users to access import settings

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869dn72ua

### :memo: Implementation
Adds a new `importSettingsAccess` permission field to `AppSettings` (stored in the `user-extended-app-settings` constant). Non-admin users listed in this field can open the settings dialog and see only the Import tab, without access to the rest of the settings.
- New `CheckCurrentUserCanAccessImportSettingsUseCase` checks: super admin, full settings access, or import-only access
- Settings gear icon is shown if the user qualifies for either access level
- Admin can configure import access via Settings → Permissions → "Access to Import Settings"

Screenshot 2026-06-15 at 08 22 43

### :fire: Testing
1. As admin, go to Settings → Permissions → enable "Access to Import Settings" and add a user/group
2. Log in as that user — gear icon is visible, clicking it shows only the Import tab
3. A user in neither list sees no gear icon

#869dn72ua


i18n/fr.po
```diff
@@ -361,6 +361,16 @@ msgstr ""
 "Les modifications sur 'Qui a accès' aux paramètres seront prises en compte "
 "après le rechargement de la page"
 
+#, fuzzy
```
eperedo: can we remove the fuzzy translations? Here, but also for PT and PT_BR

src/webapp/pages/Router.tsx
```diff
@@ -37,14 +37,20 @@ const TabWrapper = ({
...
     const [reloadKey, reload] = useReload();
 
     React.useEffect(() => {
         compositionRoot.users.checkCurrentUserCanAccessSettings().run(setCurrentUserHasAccessToSettings, console.error);
+        compositionRoot.users
+            .checkCurrentUserCanAccessImportSettings()
+            .run(setCurrentUserHasAccessToImportSettings, console.error);
     }, [compositionRoot.users]);
 
+    const showSettingsIcon = currentUserHasAccessToSettings || currentUserHasAccessToImportSettings;
```
eperedo: I'm not sure about this change. This means that users who have access to the import action can now open the 'settings' modal and configure any permission. This was not part of the task, so maybe we should stick to the previous logic?

src/domain/usecases/CheckCurrentUserCanAccessImportSettingsUseCase.ts
```diff
@@ -0,0 +1,27 @@
...
+        return Future.joinObj({
+            user: this.userRepository.getCurrent(),
+            appSettings: this.appSettingsRepository.get(),
+        }).map(({ user, appSettings }) => {
+            const access = {
+                userId: getId(user),
+                userGroupIds: user.userGroups.map(getId),
+            };
+
+            return (
```
eperedo: The whole usecase is ok, but this is almost the same logic we have in `CheckCurrentUserCanAccessSettingsUseCase.ts`. I think we can have only one usecase and instead returning a boolean we can return a new type:

```ts
type UserSettingsAccess = {
  accessToSettings: boolean;
  accessToImport: boolean;
}
```

and now we can centralize every access logic in this usecase so in the future we don't need one usecase for every new setting and also we save hitting the server with the same two requests to get the current user and the appsettings. Does this make sense? What do you think?

deeonwuli: Makes sense 👍 .

src/webapp/pages/Router.tsx
deeonwuli: Users with access to import actions only can open the settings modal and only configure the import settings, not all other permissions. See [SettingsDialogModal L210](https://github.com/EyeSeeTea/user-extended/blob/386938dc40d497f8f069b89ba45aaad900634d08/src/webapp/components/settings-dialog-modal/SettingsDialogModal.tsx#L210).

